PR #1, opened 2016-12-21
Add dial configuration

jsimonetti:
This will add a configuration to netlink.Dial which allows to
netlink groups during bind.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -38,9 +38,6 @@ type osConn interface {
 // Config specifies optional configuration for Conn.  If config is nil, a default
 // configuration will be used.
 func Dial(proto int, config *Config) (*Conn, error) {
-	if config == nil {
-		config = &Config{}
-	}
 	// Use OS-specific dial() to create osConn
 	c, err := dial(proto, config)
 	if err != nil {
--- a/conn.go
+++ b/conn.go
@@ -35,9 +35,12 @@ type osConn interface {
 }
 
 // Dial dials a connection to netlink, using the specified protocol number.
-func Dial(proto int) (*Conn, error) {
+func Dial(proto int, config *Config) (*Conn, error) {
+	if config == nil {
+		config = &Config{}
+	}
 	// Use OS-specific dial() to create osConn
-	c, err := dial(proto)
+	c, err := dial(proto, config)
 	if err != nil {
 		return nil, err
 	}
@@ -164,4 +167,9 @@ func Validate(request Message, replies []Message) error {
 	}
 
 	return nil
+}
+
+// Config contains options for the netlink socket
+type Config struct {
+	NLGroups uint32 // netlink multicast groups
 }
--- a/conn.go
+++ b/conn.go
@@ -35,6 +35,8 @@ type osConn interface {
 }
 
 // Dial dials a connection to netlink, using the specified protocol number.
+// Config specifies optional configuration for Conn.  If config is nil, a default
+// configuration will be used.
 func Dial(proto int, config *Config) (*Conn, error) {
 	if config == nil {
 		config = &Config{}
@@ -169,7 +171,9 @@ func Validate(request Message, replies []Message) error {
 	return nil
 }
 
-// Config contains options for the netlink socket
+// Config contains options for a Conn.
 type Config struct {
-	NLGroups uint32 // netlink multicast groups
+	// Groups is a bitmask which specifies multicast groups. If set to 0,
+	// no multicast group subscriptions will be made.
+	Groups uint32
 }
--- a/conn_linux_test.go
+++ b/conn_linux_test.go
@@ -294,9 +294,9 @@ func TestLinuxConnConfig(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			_, s := testLinuxConn(t, tt.config)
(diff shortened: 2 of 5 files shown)


jsimonetti:
A sample rtnetlink showing it works:
https://gist.github.com/jsimonetti/a6acaa5c6b5489ad5df8b92e23b7aaac

conn.go
```diff
@@ -35,9 +35,12 @@ type osConn interface {
 }
 
 // Dial dials a connection to netlink, using the specified protocol number.
```
mdlayher: Let's expand this comment to talk about `*Config`:

```
// Dial dials a connection to netlink, using the specified protocol number.
// Config specifies optional configuration for Conn.  If config is nil, a default
// configuration will be used.
```

conn.go
```diff
@@ -165,3 +168,8 @@ func Validate(request Message, replies []Message) error {
 
 	return nil
 }
+
+// Config contains options for the netlink socket
```
mdlayher: `s/the netlink socket/a Conn/`.

Add a period as well.

conn.go
```diff
@@ -165,3 +168,8 @@ func Validate(request Message, replies []Message) error {
...
+type Config struct {
+	NLGroups uint32 // netlink multicast groups
```
mdlayher: How about just `Groups`?  I have been dropping `NL` prefixes since this package is referred to as `netlink` externally.

Also, let's move the comment above the field and have it say something like

```
// Groups is a bitmask which specifies multicast groups.  If set to 0,
// no multicast group subscriptions will be made.
```

conn_others.go
```diff
@@ -20,7 +20,7 @@ var _ osConn = &conn{}
 type conn struct{}
 
 // dial is the entry point for Dial.  dial always returns an error.
-func dial(family int) (*conn, error) {
+func dial(family int, config *SockOpts) (*conn, error) {
```
mdlayher: This won't build yeah?  Should be `*Config`.  I'll see if I can set up OSX builds in Travis to check the "others" piece too.

jsimonetti: Oops, i used that name earlier. It should be `*Config` indeed.

jsimonetti:
Ack to the other comments.

With regard to the test, that sounds good. I Will add them.

jsimonetti:
Ready for 2nd review. I think travis is stuck. It appears to work in the first pass.

conn_linux_test.go
```diff
@@ -269,9 +269,51 @@ func TestLinuxValidate(t *testing.T) {
...
+
+		})
+	}
+
+}
+
+func testLinuxConn(t *testing.T, config *Config) (*conn, *testSocket) {
 	s := &testSocket{}
-	c, err := bind(s)
+	if config == nil {
```
mdlayher: The problem with this is that right now, our test is testing `testLinuxConn`'s version of a `Conn` instead of `Dial`'s.  I think the most expedient approach is to push the `nil` check into `bind()` and handle it per-OS, unfortunately.  This being said, I doubt this package will ever work on other platforms so that's probably okay.

conn.go
mdlayher: This is resolved.

conn_linux_test.go
```diff
@@ -269,9 +269,51 @@ func TestLinuxValidate(t *testing.T) {
 	}
 }
 
-func testLinuxConn(t *testing.T) (*conn, *testSocket) {
+func TestLinuxConnConfig(t *testing.T) {
+
```
mdlayher: I know this is silly, but I'm not a fan of whitespace at the start or end of function bodies.  Mind removing this one, lines 298, 305, and 310?  Sorry.  Just a personal preference.

conn_linux_test.go
```diff
@@ -269,9 +269,51 @@ func TestLinuxValidate(t *testing.T) {
...
+	}{
+		{
+			name:   "Default Config",
+			config: &Config{},
+			groups: 0x0,
+		},
+		{
+			name:   "Config with Groups RTMGRP_IPV4_IFADDR",
+			config: &Config{Groups: 0x10},
+			groups: 16,
```
mdlayher: Do `0x10` to match the config.

conn_linux_test.go
```diff
@@ -269,9 +269,51 @@ func TestLinuxValidate(t *testing.T) {
...
+		},
+		{
+			name:   "Config with Groups RTMGRP_IPV4_IFADDR | RTMGRP_IPV4_ROUTE",
+			config: &Config{Groups: 0x10 | 0x40},
+			groups: 80,
```
mdlayher: Do `0x50` to match the config.

mdlayher: https://play.golang.org/p/XR0TaQzVly

conn_linux_test.go
jsimonetti: I pushed the nil check down to `bind()`. Unfortunately I don't see a way to check if the resulting `Conn` actually uses the group. The test only checks to see whether the bind sfails or not with the groups set. We don't have access to the `syscall.SockaddrNetlink` from a `Conn`.

mdlayher: We can figure out a decent integration test at a later time. Merged this for now as it seems fine and has good enough test coverage.

PR #2, opened 2016-12-22
travis: test on OSX and Linux

mdlayher:
This way we can verify that `*_others.go` isn't broken when changes are made.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: go
 go:
   - 1.x
+os:
+  - linux
+  - osx
 before_install:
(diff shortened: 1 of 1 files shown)


jsimonetti:
Perhaps we need a test for others platforms?
Should be straightforward, right?

mdlayher:
Yep, I have this at the moment: https://github.com/mdlayher/netlink/blob/master/conn_others_test.go.

jsimonetti:
Yes, sorry. I am blind. Disregard my previous comment.

PR #10, opened 2016-12-23
Add netlink multicast test

jsimonetti:
Add a proper CI test for netlink groups.


conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+func TestLinuxNetlinkMulticast(t *testing.T) {
+	cfg := &Config{
+		Groups: 0x1, // RTMGRP_LINK
+	}
+
+	c, err := Dial(0, cfg) // dials NETLINK_ROUTE
+	if err != nil {
+		t.Fatalf("failed to dial: %v", err)
+
```
mdlayher: Yank this empty line.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	in := make(chan []Message)
+
+	// routine for receiving any messages
+	recv := func() {
+		data, er := c.Receive()
```
mdlayher: Let's stick with `err` for consistency.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+		if er != nil {
+			fmt.Printf("error in receive: %s\n", err)
```
mdlayher: `t.Fatalf`, or maybe even `panic(fmt.Sprintf())`.  Not sure what happens when you `t.Fatal` in a goroutine.

mdlayher: Don't need newlines when using `t.Fatal`.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+	def := sudoIfCreate(t, "test0")
+	defer def()
+
+	timeout := time.After(5 * time.Second)
+	var got []Message
+	select {
+	case got = <-in:
+		break
+	case <-timeout:
+		t.Fatal("did not receive any messages after 5 seconds\n")
```
mdlayher: Don't need newlines when using `t.Fatal`.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+		break
+	}
+	t.Logf("received netlink data: %#v\n", got)
+}
+
+func sudoIfCreate(t *testing.T, ifName string) func() {
+	cmd := exec.Command("sudo", "ip", "tuntap", "add", ifName, "mode", "tun")
+	err := cmd.Start()
+	if err != nil {
+		t.Fatalf("error creating tuntap device: %s\n", err)
```
mdlayher: Don't need newlines when using `t.Fatal`.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	cmd.Wait()
+
+	return func() {
+		cmd := exec.Command("sudo", "ip", "link", "del", ifName)
+		err := cmd.Start()
+		if err != nil {
+			t.Fatalf("error removing tuntap device: %s\n", err)
```
mdlayher: Don't need newlines when using `t.Fatal`.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		cmd.Wait()
```
mdlayher: Might as well just check this error too.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,67 @@
...
+	timeout := time.After(5 * time.Second)
+	var got []Message
+	select {
+	case got = <-in:
+		break
+	case <-timeout:
+		t.Fatal("did not receive any messages after 5 seconds\n")
+		break
+	}
+	t.Logf("received netlink data: %#v\n", got)
```
mdlayher: Obviously this isn't a `rtnetlink` package, but are there any kind of header values we can inspect quickly here just to determine that this is a valid response and not something bogus?

jsimonetti:
I have updated the PR with the requested style changes.

I also added a check in the received message.
The message should be a ifinfomsg with the interface name equal to the interface we created during the test.
To go much further, we probably have to decode the message data.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	interf := []byte(ifName)
+	want := make([]uint8, len(ifName))
+	copy(want, interf[:])
+
+	got := make([]uint8, len(ifName))
+	copy(got, data[0].Data[20:len(ifName)+20])
+
+	if !reflect.DeepEqual(want, got) {
+		t.Fatalf("received message doesn not mention ifName %s", ifName)
```
mdlayher: *does, and let's go with `%q` so any weird whitespace or whatever is easily spotted if this fails.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+	def := sudoIfCreate(t, ifName)
+	defer def()
+
+	timeout := time.After(5 * time.Second)
+	var data []Message
+	select {
+	case data = <-in:
+		break
+	case <-timeout:
+		t.Fatal("did not receive any messages after 5 seconds")
```
mdlayher: `panic()` is nice in the case of timeouts because it'll give a stack trace showing where the goroutines are hung up.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,81 @@
+//+build integration,linux
+
+package netlink
+
+import (
+	"fmt"
+	"os/exec"
+	"testing"
+	"time"
+
```
mdlayher: Weird, mind removing this?

conn_linux_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func sudoIfCreate(t *testing.T, ifName string) func() {
+	cmd := exec.Command("sudo", "ip", "tuntap", "add", ifName, "mode", "tun")
+	err := cmd.Start()
+	if err != nil {
+		t.Fatalf("error creating tuntap device: %s", err)
+		return func() {}
+	}
+	cmd.Wait()
```
mdlayher: Check the error.

conn_linux_integration_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return func() {
+		cmd := exec.Command("sudo", "ip", "link", "del", ifName)
+		err := cmd.Start()
+		if err != nil {
+			panic(fmt.Sprintf("error removing tuntap device: %s", err))
+		}
+		cmd.Wait()
```
mdlayher: Check the error.

jsimonetti:
One addition commit to address the requested changes.
Feel free to squash this at merge time btw.

PR #49, opened 2017-03-11
Change dial() signature in conn_others following re-factoring

ti-mo:
Noticed a small issue running `go get` on mac. This should fix it. :)

Cheers


mdlayher:
Thanks for this. I had Travis OSX builds but they sometimes took an hour or more to start, so I got tired of waiting.

I'll try to remember to check locally next time too.

mdlayher:
Apparently I can't merge from the ForkHub app so I'll just do it later.

PR #56, opened 2017-03-17
Attribute helpers for nesting, endianness and a mask over .Type

ti-mo:
Hi, I've added three helpers to `Attribute` including some tests. These are all exported to allow eg. NetFilter to make use of them, which should be their main purpose IMO. They're not implemented as struct members to avoid having to change (and bloat) the struct.

- IsNested() - masks leftmost bit
- IsNetByteOrder() - masks second leftmost bit
- GetType() - masks all other bits from the Type

I have benched using pointer receivers for these, but there was no perceivable performance difference. Let me know what you think!

T

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 
 	"github.com/mdlayher/netlink/nlenc"
-	"golang.org/x/sys/unix"
 )
 
 var (
@@ -25,25 +24,26 @@ type Attribute struct {
 	Data []byte
 }
 
-// IsNested interprets the Type field of the Attribute to determine whether it's
-// a nested attribute or not. This facilitates recursive parsing of the attribute.
-// This is the leftmost bit in Type. Mutually exclusive with IsNetByteOrder().
+// #define NLA_F_NESTED
+const nlaNested uint16 = 0x8000
+// #define NLA_F_NET_BYTE_ORDER
+const nlaNetByteOrder uint16 = 0x4000
+// Masks all bits except for Nested and NetByteOrder
+const nlaTypeMask = ^(nlaNested | nlaNetByteOrder)
+
+// IsNested determines if an Attribute's data contains additional nested attributes.
 func (a Attribute) IsNested() bool {
-	return (a.Type & unix.NLA_F_NESTED) > 0
+	return (a.Type & nlaNested) > 0
 }
 
 // IsNetByteOrder interprets the Type field of the Attribute to determine whether the attribute
 // is big endian (net byte order) or not.
-// This is the second bit from the left in Type. Mutually exclusive with IsNested().
 func (a Attribute) IsNetByteOrder() bool {
-	return (a.Type & unix.NLA_F_NET_BYTEORDER) > 0
+	return (a.Type & nlaNetByteOrder) > 0
 }
 
-var nlaTypeMask = ^uint16(unix.NLA_F_NESTED | unix.NLA_F_NET_BYTEORDER)
-
 // GetType masks the Attribute's Type with the bits that are NOT
-// reserved for NLA_F_NESTED and NLA_F_NET_BYTEORDER.
-// Only the 14 rightmost bits will be used.
+// reserved for nlaNested and nlaNetByteOrder.
 func (a Attribute) GetType() uint16 {
 	return a.Type & nlaTypeMask
 }
--- a/attribute.go
+++ b/attribute.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 
 	"github.com/mdlayher/netlink/nlenc"
+	"golang.org/x/sys/unix"
 )
 
 var (
@@ -22,6 +23,29 @@ type Attribute struct {
 
 	// An arbitrary payload which is specified by Type.
 	Data []byte
+}
+
+// Interprets the Type field of the Attribute to determine whether it's
+// a nested attribute or not. This facilitates recursive parsing of the attribute.
+// This is the leftmost bit in Type. Mutually exclusive with IsNetByteOrder().
+func (a Attribute) IsNested() bool {
+	return (a.Type & unix.NLA_F_NESTED) > 0
+}
+
+// Interprets the Type field of the Attribute to determine whether the attribute
+// is big endian (net byte order) or not.
+// This is the second bit from the left in Type. Mutually exclusive with IsNested().
+func (a Attribute) IsNetByteOrder() bool {
+	return (a.Type & unix.NLA_F_NET_BYTEORDER) > 0
+}
+
+const NLA_TYPE_MASK = ^uint16(unix.NLA_F_NESTED | unix.NLA_F_NET_BYTEORDER)
+
+// Mask the Attribute's Type with the bits that are NOT
+// reserved for NLA_F_NESTED and NLA_F_NET_BYTEORDER.
+// Only the 14 rightmost bits will be used.
+func (a Attribute) GetType() uint16 {
+	return a.Type & NLA_TYPE_MASK
 }
 
 // MarshalBinary marshals an Attribute into a byte slice.
--- a/attribute.go
+++ b/attribute.go
@@ -9,6 +9,9 @@ import (
 var (
 	// errInvalidAttribute specifies if an Attribute's length is incorrect.
 	errInvalidAttribute = errors.New("invalid attribute; length too short or too large")
+	// errInvalidAttributeFlags specifies if an Attribute's flag configuration is invalid.
+	// From a comment in Linux/include/uapi/linux/netlink.h, Nested and NetByteOrder are mutually exclusive.
+	errInvalidAttributeFlags = errors.New("invalid attribute; type cannot have both nested and net byte order flags")
 )
 
 // An Attribute is a netlink attribute.  Attributes are packed and unpacked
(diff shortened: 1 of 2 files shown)


attribute.go
```diff
@@ -24,6 +25,29 @@ type Attribute struct {
 	Data []byte
 }
 
+// IsNested interprets the Type field of the Attribute to determine whether it's
+// a nested attribute or not. This facilitates recursive parsing of the attribute.
+// This is the leftmost bit in Type. Mutually exclusive with IsNetByteOrder().
+func (a Attribute) IsNested() bool {
+	return (a.Type & unix.NLA_F_NESTED) > 0
```
mdlayher: Using `unix` outside of `*_linux.go` files means that it won't build on non-UNIX machines.  Need to make a copy of these constants locally.

ti-mo: I see, that's why I saw them duplicated across the package. Got it!

attribute.go
```diff
@@ -24,6 +25,29 @@ type Attribute struct {
 	Data []byte
 }
 
+// IsNested interprets the Type field of the Attribute to determine whether it's
```
mdlayher: I think this can be shortened to:

```
// IsNested determines if an Attribute's data contains additional nested attributes.
```

The rest is an implementation detail IMO.

attribute.go
```diff
@@ -24,6 +25,29 @@ type Attribute struct {
...
+// a nested attribute or not. This facilitates recursive parsing of the attribute.
+// This is the leftmost bit in Type. Mutually exclusive with IsNetByteOrder().
+func (a Attribute) IsNested() bool {
+	return (a.Type & unix.NLA_F_NESTED) > 0
+}
+
+// IsNetByteOrder interprets the Type field of the Attribute to determine whether the attribute
```
mdlayher: Out of curiosity, what part of the attribute would be big endian? Is it just the data portion?

attribute.go
```diff
@@ -24,6 +25,29 @@ type Attribute struct {
...
+// is big endian (net byte order) or not.
+// This is the second bit from the left in Type. Mutually exclusive with IsNested().
+func (a Attribute) IsNetByteOrder() bool {
+	return (a.Type & unix.NLA_F_NET_BYTEORDER) > 0
+}
+
+var nlaTypeMask = ^uint16(unix.NLA_F_NESTED | unix.NLA_F_NET_BYTEORDER)
```
mdlayher: `const`, and add a doc comment here too for sure.

attribute.go
ti-mo: I've also been trying to figure this out, but it seems like this is only consumed by IPset in userspace. Most of the uses I've seen are to throw errors when the message received is not big endian. Will definitely do some further research.

ti-mo: @mdlayher Your initial guess was correct, this flag doesn't indicate a change to the endianness of the header itself. It only seems to be used in IPSet. Following [an example call chain `nla_put_net16`](http://lxr.free-electrons.com/source/net/netfilter/ipset/ip_set_hash_ipport.c#L69) all the way down to [`__nla_put`](http://lxr.free-electrons.com/source/lib/nlattr.c#L482):

```
 void __nla_put(struct sk_buff *skb, int attrtype, int attrlen,
                              const void *data)
{
        struct nlattr *nla;

        nla = __nla_reserve(skb, attrtype, attrlen);
        memcpy(nla_data(nla), data, attrlen);
}
```

In the example above, `attrtype` is `IPSET_ATTR_PORT` OR'ed with `NLA_F_NET_BYTEORDER`. `data` is a `(struct hash_ipport4_elem)->data`, which is supposed to be a big-endian representation of a 16-bit port number. `attrtype` has the big-endian bit enabled to indicate this. Supporting this bit would allow us to successfully interpret IPSet in a NetFilter package. 👍

attribute.go
```diff
@@ -24,6 +24,30 @@ type Attribute struct {
...
+
+// IsNetByteOrder interprets the Type field of the Attribute to determine whether the attribute
+// is big endian (net byte order) or not.
+func (a Attribute) IsNetByteOrder() bool {
+	return (a.Type & nlaNetByteOrder) > 0
+}
+
+// GetType masks the Attribute's Type with the bits that are NOT
+// reserved for nlaNested and nlaNetByteOrder.
+func (a Attribute) GetType() uint16 {
```
mdlayher: For ease of use, I think it might be a good idea to just store this in `a.Type` directly. And then perhaps it'd be reasonable to just add struct fields `Nested bool` and `NetByteOrder bool` to Attribute, since we aren't relying on its size for any reason anyway because it contains an arbitrary byte slice of data.

attribute_test.go
```diff
@@ -367,3 +367,70 @@ func TestUnmarshalAttributes(t *testing.T) {
 		})
 	}
 }
+
+func TestAttribute_GetType(t *testing.T) {
+	// Small test bed of values that map to
+	// counterparts with the two leftmost bits zeroed out.
+	types := map[uint16]uint16{
```
mdlayher: Nit: let's go with slices for these tests so the test order is deterministic.

attribute_test.go
```diff
@@ -367,3 +367,70 @@ func TestUnmarshalAttributes(t *testing.T) {
...
+		0xAAAA: 0x2AAA,
+		0xFFFF: 0x3FFF,
+		0x2AAA: 0x2AAA,
+		0x0FFF: 0x0FFF,
+		0x0: 0x0,
+	}
+
+	for input, want := range types {
+		test := Attribute{
```
mdlayher: Fire off a subtest with the hex representation of the type:

```go
for _, tt := range tests {
    t.Run(fmt.Sprintf("0x%04x", tt.attr), func(t *testing.T) {
        // ...
    })
}
```

attribute_test.go
```diff
@@ -367,3 +367,70 @@ func TestUnmarshalAttributes(t *testing.T) {
...
+		}
+
+		if got := test.GetType(); got != want {
+			t.Fatalf("unexpected result:\n- want: %.16b\n- got: %.16b",
+				want, got)
+		}
+	}
+}
+
+func TestAttribute_IsNested(t *testing.T) {
```
mdlayher: Same comments for all tests.

attribute.go
mdlayher: Regarding bloating the struct: I'm not too worried about it and think the usability improvement beats adding more methods to the API.

attribute_test.go
ti-mo: Sorry for breaking convention, went with extending `TestUnmarshalAttributes`

attribute.go
```diff
@@ -22,8 +22,21 @@ type Attribute struct {
 
 	// An arbitrary payload which is specified by Type.
 	Data []byte
+
+	// Whether the attribute contains nested attributes
```
mdlayher: Add a period.

attribute.go
```diff
@@ -22,8 +22,21 @@ type Attribute struct {
...
+	Nested bool
+
+	// Endianness of the payload
```
mdlayher: `// Whether the attribute's data is in network byte order or native byte order.`, or something like that.

attribute.go
```diff
@@ -22,8 +22,21 @@ type Attribute struct {
...
+	NetByteOrder bool
 }
 
+// #define NLA_F_NESTED
+const nlaNested uint16 = 0x8000
+// #define NLA_F_NET_BYTE_ORDER
+const nlaNetByteOrder uint16 = 0x4000
+// Masks all bits except for Nested and NetByteOrder
```
mdlayher: Add a period.

attribute.go
```diff
@@ -33,7 +46,16 @@ func (a Attribute) MarshalBinary() ([]byte, error) {
 	b := make([]byte, nlaAlign(int(a.Length)))
 
 	nlenc.PutUint16(b[0:2], a.Length)
-	nlenc.PutUint16(b[2:4], a.Type)
+
+	switch {
```
mdlayher: Add a case at the top that checks for both at the same time.

attribute_test.go
```diff
@@ -346,6 +346,68 @@ func TestUnmarshalAttributes(t *testing.T) {
 				},
 			},
 		},
+		{
```
mdlayher: Add the same tests in `TestMarshalAttributes`.

attribute.go
```diff
@@ -22,18 +25,44 @@ type Attribute struct {
 
 	// An arbitrary payload which is specified by Type.
 	Data []byte
+
+	// Whether the attribute contains nested attributes.
```
mdlayher: Let's specify "attribute's data" instead of just "attribute".

attribute.go
```diff
@@ -22,18 +25,44 @@ type Attribute struct {
...
+	Nested bool
+
+	// Whether the attribute is in network (true) or native (false) byte order.
```
mdlayher: Same: "attribute's data" instead of just "attribute".

attribute.go
```diff
@@ -22,18 +25,44 @@ type Attribute struct {
...
+	NetByteOrder bool
 }
 
+// #define NLA_F_NESTED
+const nlaNested uint16 = 0x8000
+// #define NLA_F_NET_BYTE_ORDER
+const nlaNetByteOrder uint16 = 0x4000
+// Masks all bits except for Nested and NetByteOrder
```
mdlayher: Add a period.

PR #60, opened 2017-03-21
Parsing nested Attributes 

ti-mo:
Following the changes in #56, this handles parsing nested Attributes.

Note that `UnmarshalAttributes` is called from `a.UnmarshalBinary`, since it does exactly what we need to generate an `[]Attribute`. If this ever becomes a problem, we can duplicate minimal functionality into `UnmarshalBinary`.

T


mdlayher:
Looks like there is a merge conflict.  Rebase?  I'll give this a go with my genetlink stuff and report back.  If attributes that contain nested attributes aren't flagged consistently, it may be better to leave it up to the caller to decide what should be unmarshaled. Plus, it would save some allocations for the cases where you don't necessarily want or need every nested attribute of every attribute.

mdlayher:
Yep, as I suspected, nl80211 at least does not flag attributes as nested consistently.  To avoid potential confusion I've decided that it is best to leave it up to the programmer to determine if attributes are nested or not, rather than trying to do it automatically.

In addition, as I said, I'm not sure the lines of code saved are worth the added cost of actually unmarshaling and allocating every single nested attribute in the cases where only one or two are really needed.

Closing.

PR #66, opened 2017-09-09
support uint8 encoding and decoding in nlenc

### LINKED ISSUE #65: nlenc: support uint8 encoding and decoding in nlenc
When we create a Data in Attribute, we usually need to encode a uintN
into a byte slice. However, as nlenc does not support uint8 encoding and
decoding, we have to create similar function for only uint8 in the
code. We would like to make our code consistent, so will be happy if
nlenc supports this.

nak3:
This patch supports uint8 encoding and decoding functions in nlenc.

Fixes #65

### DIFF
--- a/nlenc/int.go
+++ b/nlenc/int.go
@@ -5,6 +5,16 @@ import (
 	"unsafe"
 )
 
+// PutUint8 encodes a uint8 into b.
+// If b is not exactly 1 byte in length, PutUint8 will panic.
+func PutUint8(b []byte, v uint8) {
+	if l := len(b); l != 1 {
+		panic(fmt.Sprintf("PutUint8: unexpected byte slice length: %d", l))
+	}
+
+	b[0] = v
+}
+
 // PutUint16 encodes a uint16 into b using the host machine's native endianness.
 // If b is not exactly 2 bytes in length, PutUint16 will panic.
 func PutUint16(b []byte, v uint16) {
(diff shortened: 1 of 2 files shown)


nlenc/int.go
```diff
@@ -5,6 +5,16 @@ import (
...
 )
 
+// PutUint8 encodes a uint8 into b.
+// If b is not exactly 1 bytes in length, PutUint8 will panic.
+func PutUint8(b []byte, v uint8) {
+	if l := len(b); l != 1 {
+		panic(fmt.Sprintf("PutUint8: unexpected byte slice length: %d", l))
+	}
+
+	*(*uint8)(unsafe.Pointer(&b[0])) = v
```
mdlayher: Can simplify to just `b[0] = v`.

nlenc/int.go
```diff
@@ -5,6 +5,16 @@ import (
 	"unsafe"
 )
 
+// PutUint8 encodes a uint8 into b.
+// If b is not exactly 1 bytes in length, PutUint8 will panic.
```
mdlayher: "1 byte", not plural for one byte.

nlenc/int.go
```diff
@@ -45,6 +55,16 @@ func PutInt32(b []byte, v int32) {
 	*(*int32)(unsafe.Pointer(&b[0])) = v
 }
 
+// Uint8 decodes a uint8 from b.
+// If b is not exactly 1 bytes in length, Uint8 will panic.
```
mdlayher: "1 byte", not plural for one byte.

nlenc/int.go
```diff
@@ -45,6 +55,16 @@ func PutInt32(b []byte, v int32) {
...
+func Uint8(b []byte) uint8 {
+	if l := len(b); l != 1 {
+		panic(fmt.Sprintf("Uint8: unexpected byte slice length: %d", l))
+	}
+
+	return *(*uint8)(unsafe.Pointer(&b[0]))
```
mdlayher: Simplify: `return b[0]`.

nlenc/int.go
```diff
@@ -85,6 +105,14 @@ func Int32(b []byte) int32 {
 	return *(*int32)(unsafe.Pointer(&b[0]))
 }
 
+// Uint8Bytes encodes a uint8 into a newly-allocated byte slice. It is a
+// shortcut for allocating a new byte slice and filling it using PutUint16.
```
mdlayher: *PutUint8

nak3:
Thank you. Updated.

PR #67, opened 2017-09-09
docs: move comments about members of Header struct

nak3:
This patch moves comments about if some values are `0` on Send()
function to members of Header struct.  These comments should be
 checked on Header struct in the GoDocs.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -106,15 +106,6 @@ func (c *Conn) Execute(m Message) ([]Message, error) {
 // Sequence, and PID fields should be set to 0, so they can be populated
 // automatically before the Message is sent.  On success, Send returns a copy
 // of the Message with all parameters populated, for later validation.
-//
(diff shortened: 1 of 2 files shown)


mdlayher:
This change is actually incorrect. The Message type does not infer any field values on its own.  If you look at the source, Conn.Send handles this.

Since the majority of users just want to send netlink messages and don't care about the actual structure, this case is handled exclusively in Conn.Send, and therefore the current documentation is correct. If a user would like to create a very specific message, we leave it up to them to craft it, validate it, and marshal it to bytes.

Basically, this makes the common use case very easy and well documented, but also provides the ability for a user to gain full control.

Thanks for the PR, but I'd prefer to keep this as is for the reasons mentioned here.

nak3:
I understand and agree. Thank you for your clarification.

mdlayher:
No problem! Thanks for your time as well.

PR #71, opened 2017-12-04
Added missing new object HeaderFlags.

brlbil:
Modified String() function. I have used the same string format as I have seen Flags printed in the strace output. Could not come up with a better one.

I have used missing Flags in a project, I could simply define them in there but I believe they belong in this package.

### DIFF
--- a/message.go
+++ b/message.go
@@ -62,6 +62,20 @@ const (
 	// HeaderFlagsDump requests that netlink return a complete list of
 	// all entries.
 	HeaderFlagsDump HeaderFlags = HeaderFlagsRoot | HeaderFlagsMatch
+
+	// Flags used to create objects.
+
+	// HeaderFlagsReplace indicates request replaces an existing matching object.
+	HeaderFlagsReplace HeaderFlags = 0x100
+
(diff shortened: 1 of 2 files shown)


mdlayher:
Seems reasonable, but I will do an in-depth review tomorrow. Out of curiosity, what's your project? Would like to see the source!

brlbil:
I am working on github.com/jsimonetti/rtnetlink which I will later use for a project, which is on the drawing board at this point. I have added RouteSerive to rtnetlink and also made some changes which are not ready to push yet. I would like to add missing tests but I could not get around to do it yet.

As an example of usage, in rtnetlink AddressService New function only netlink.HeaderFlagsRequest flag is used, this seems to be wrong needs to replaced with netlink.HeaderFlagsRequest | headerFlagCreate | netlink.HeaderFlagsAcknowledge | headerFlagExcl. Obviously, headerFlagCreate and headerFlagExcl are private variables at this point.

mdlayher:
LGTM!

mdlayher:
Looking forward to checking out your project!

PR #72, opened 2017-12-14
netlink: by default, lock OS thread internally

mdlayher:
/cc @squeed


squeed:
I think you can safely unlock when you're done with a particular connection, because if the netlink socket is closed, you won't have the problem you were trying to avoid. So you can probably safely unlock and preserve go1.9 compat.

Issue #74, opened 2018-03-26
Why not change implemention of receive() func


HeChuanXUPT:
Hi,
https://github.com/mdlayher/netlink/blob/756e798fb38fac19fb2234d3acc32e902bc1af44/conn.go#L179

The recursive implement for this func is not fast when large number of tcp sockets(test in more then 40,000)
Can I change to loop to receive msgs?

PR #75, opened 2018-03-26
receive iterative implement instead recursive

HeChuanXUPT:
#74


mdlayher:
This looks good from a code simplification perspective.  Will merge.

PR #76, opened 2018-05-10
update TestLinuxConnIntegration for recent kernel versions

pgier:
Recent kernel versions return the NLM_F_ROOT (0x100) header flag.

I'm not sure if this is the right fix for this, but this test currently fails on recent distributions (Fedora 27 and Ubuntu 18.04) because the header section of the message contains the NLM_F_ROOT flag (256/0x100).


mdlayher:
Interesting!  Seems good to me though.

PR #80, opened 2018-05-16
netlink: remove Nested and NetByteOrder from Attribute

mdlayher:
I haven't seen a netlink message yet that actually uses these, and since nested attributes in particular appear all the time without this flag being set, it doesn't seem to provide any added value.

For the time being, let's remove these.


ti-mo:
@mdlayher It's a bummer this is gone, Netfilter uses them extensively, too bad you forgot about this. Didn't make sense to have it without the `Children []Attribute` in the same struct anyway, it ended up being a bit annoying to embed `netlink.Attribute` and having to go through `nfa.Attribute.Nested` all the time.

PR #88, opened 2018-07-11
fixes building on non-linux systems

viert:
SendMessages() added to non-linux implementation of conn to match Socket interface. Some packages depending on netlink fail to build due to this since SendMessages method was merged in May 2018

### DIFF
--- a/conn_others.go
+++ b/conn_others.go
@@ -31,6 +31,11 @@ func (c *conn) Send(m Message) error {
 	return errUnimplemented
 }
 
+// SendMessages always returns an error.
+func (c *conn) SendMessages(m []Message) error {
+	return errUnimplemented
+}
+
 // Receive always returns an error.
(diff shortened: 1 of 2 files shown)


conn_others.go
```diff
@@ -31,6 +31,11 @@ func (c *conn) Send(m Message) error {
 	return errUnimplemented
 }
 
+// SendMessages always returns an error
```
mdlayher: Nit: add a period.

conn_others_test.go
```diff
@@ -25,6 +25,11 @@ func TestOthersConnUnimplemented(t *testing.T) {
 			want, got)
 	}
 
+	if got := c.SendMessages([]Message{Message{}}); want != got {
```
mdlayher: Nit: just pass `nil` for simplicity since it'll always return an error anyway.

viert: yes, agree

PR #90, opened 2018-07-17
DetachBPF

florianl:
Updated PR for the implementation of DetachBPF.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -310,11 +310,11 @@ func (c *Conn) LeaveGroup(group uint32) error {
 	return gc.LeaveGroup(group)
 }
 
-// A bpfSetter is a Socket that supports setting BPF filters.
+// A bpfSetter is a Socket that supports setting and removing BPF filters.
 type bpfSetter interface {
 	Socket
 	bpf.Setter
-	DetachBPF() error
+	RemoveBPF() error
 }
 
 // SetBPF attaches an assembled BPF program to a Conn.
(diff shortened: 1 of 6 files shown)


conn.go
```diff
@@ -314,6 +314,7 @@ func (c *Conn) LeaveGroup(group uint32) error {
 type bpfSetter interface {
 	Socket
 	bpf.Setter
+	DetachBPF() error
```
mdlayher: Let's go with "RemoveBPF" to more closely match the existing "SetBPF" terminology used in this and other Go packages.

Please update the interface comment to say "setting and removing BPF filters.".

conn_linux.go
```diff
@@ -226,6 +226,16 @@ func (c *conn) SetBPF(filter []bpf.RawInstruction) error {
 	)
 }
 
+// DetachBPF removes a BPF filter from a conn.
+func (c *conn) DetachBPF() error {
+	var dummy int
```
mdlayher: This should be `uint32`, and let's leave a comment explaining that the option value is ignored.

conn_linux.go
```diff
@@ -226,6 +226,16 @@ func (c *conn) SetBPF(filter []bpf.RawInstruction) error {
...
+	return c.s.SetSockopt(
+		unix.SOL_SOCKET,
+		unix.SO_DETACH_FILTER,
+		unsafe.Pointer(&dummy),
+		uint32(unsafe.Sizeof(dummy)))
```
mdlayher: Add a newline before the final brace so it lines up nicely at the first level of indentation.

conn_linux.go
```diff
@@ -424,3 +434,12 @@ func (s *sysSocket) SetSockopt(level, name int, v unsafe.Pointer, l uint32) erro
 
 	return err
 }
+
+func (s *sysSocket) DetachBFP() error {
```
mdlayher: This is never used, remove.

conn_linux_test.go
```diff
@@ -683,3 +683,14 @@ func (s *testSocket) SetSockopt(level, name int, v unsafe.Pointer, l uint32) err
 
 	return nil
 }
+
+func (s *testSocket) DetachBPF() error {
```
mdlayher: This is never used, remove.

conn_linux.go
```diff
@@ -226,6 +226,19 @@ func (c *conn) SetBPF(filter []bpf.RawInstruction) error {
 	)
 }
 
+// RemoveBPF removes a BPF filter from a conn.
+func (c *conn) RemoveBPF() error {
+	// dummy is ignored as argument to SO_DETACH_FILTER
+	// but SetSockopt requires it as an argument
+	var dummy int
```
mdlayher: `uint32` please for consistency across architectures.

Issue #92, opened 2018-08-31
Closing a Netlink socket with ongoing Receive() call


ti-mo:
Hi Matt

I'm trying to cleanly terminate a Netlink socket listening on a multicast group.

```go
func main() {
	conn, err := netlink.Dial(0xc, nil) // NETLINK_NETFILTER
	if err != nil {
		log.Fatal(err)
	}

	err = conn.JoinGroup(10) // Non-existent multicast group (Receive() will never return)
	if err != nil {
		log.Fatal(err)
	}

	recvChan := make(chan []netlink.Message)

	// Start a blocking receive
	go func() {
		recv, err := conn.Receive()
		if err != nil {
			log.Fatal(err)
		}
		recvChan <- recv
	}()

	// Make sure goroutine enters Receive()
	// time.Sleep(50 * time.Millisecond)

	err = conn.Close()
	if err != nil {
		log.Fatal(err)
	}

	log.Println(<-recvChan)
}
```

This uncovers two issues:

1. A race condition: running this program multiple times in sequence yields different results. One is a panic:

```
panic: send on closed channel
goroutine 4 [running]:
github.com/mdlayher/netlink.(*sysSocket).do(0xc4200bc0a0, 0xc4200f6000)
    /home/timo/go/src/github.com/mdlayher/netlink/conn_linux.go:356 +0xa8
```

.. that can be avoided by closing the channel after the WaitGroup merges. This causes a deadlock in the test suite, so this might invalidate some assumptions you've made in other places.

```go
func (s *sysSocket) Close() error {
	...
	s.wg.Wait()
	close(s.funcC)
```

2. `Receive()` is non-interruptible. (uncomment the `Sleep()` in the example)

If there's no valid Netlink message coming in (like in the example above), `Receive()` will hang forever. I've tried closing the fd from the main thread by executing `unix.Close()` directly instead of queueing it with `s.do()`, but this (expectedly) doesn't interrupt the syscall either. It will only yield when a message is read from the socket.

I know this is not directly an issue with the library itself, but it would be pleasant to have a solution or workaround. Is there anything you could think of? A quick search hinted that syscalls can be interrupted by signals, not sure if this is something we could implement.

mdlayher:
Thanks for looking into this. I'll see what I can do. I do have an issue open to see if we can integrate with the runtime network poller properly, but it may be difficult due to the need for certain system calls.

Either way we should fix the closing synchronization.

florianl:
[...]

PR #93, opened 2018-09-03
Avoid panic on closed socket

florianl:
This PR adds an additional check, to avoid the panic mentioned in issue #92 by @ti-mo


florianl:
For Mac OSX:
```
The command "go get -d ./..." failed and exited with 1 during .
```

ti-mo:
I don't think this is necessary.. `close(s.funcC)` should be moved to execute after `s.wg.Wait()` instead.

mdlayher:
This won't quite work because of the possibility of a data race on done. I'll look more into this when I am home tomorrow.

PR #95, opened 2018-09-05
netlink: add AttributeEncoder type

terinjokes:
The AttributeDecoder type makes it easy to quickly and safely decode
fields from netlink messages. Users, however, were left to use the
low-level functions to encode messages.

This CL introduces the AttributeEncoder type, which parallels the
AttributeDecoder API, making it just as easy to encode netlink messages,
including nested attributes.

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -296,3 +296,141 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
 		ad.err = err
 	}
 }
+
+// An AttributeEncoder provides a safe way to encode attributes.
+//
+// It is recommended to use an AttributeEncoder where possible instead of
+// calling MarshalAttributes or using package nlenc directly.
+//
+// Errors from intermediate encoding steps are returned in the call to
+// Encode.
+type AttributeEncoder struct {
+	// ByteOrder defines a specific byte order to use when processing integer
+	// attributes.  ByteOrder should be set immediately after creating the
+	// AttributeEncoder: before any attributes are encoded.
+	//
+	// If not set, the native byte order will be used.
+	ByteOrder binary.ByteOrder
+
+	attrs []Attribute
+	err   error
+}
+
+// NewAttributeEncoder creates an AttributeEncoder that encodes Attributes.
+func NewAttributeEncoder() *AttributeEncoder {
+	return &AttributeEncoder{
+		ByteOrder: nlenc.NativeEndian(),
+	}
+}
+
+// Uint8 encodes the uint8 data for the specified field.
+func (ae *AttributeEncoder) Uint8(t uint16, v uint8) {
+	if ae.err != nil {
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: []byte{v},
+	})
+}
+
+// Uint16 encodes the uint16 data for the specified field.
+func (ae *AttributeEncoder) Uint16(t uint16, v uint16) {
+	if ae.err != nil {
+		return
+	}
+
+	b := make([]byte, 2)
+	ae.ByteOrder.PutUint16(b, v)
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Uint32 encodes the uint32 data for the specified field.
+func (ae *AttributeEncoder) Uint32(t uint16, v uint32) {
+	if ae.err != nil {
+		return
+	}
+
+	b := make([]byte, 4)
+	ae.ByteOrder.PutUint32(b, v)
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Uint64 encodes the uint64 data for the specified field.
+func (ae *AttributeEncoder) Uint64(t uint16, v uint64) {
+	if ae.err != nil {
+		return
+	}
+
+	b := make([]byte, 8)
+	ae.ByteOrder.PutUint64(b, v)
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// String encodes string s as a null-terminated string for the specified field.
+func (ae *AttributeEncoder) String(t uint16, s string) {
+	if ae.err != nil {
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: nlenc.Bytes(s),
+	})
+}
+
+// Bytes embeds raw byte data in the specified field.
+func (ae *AttributeEncoder) Bytes(t uint16, b []byte) {
+	if ae.err != nil {
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Do is a general purpose function to encode arbitrary data into a field.
+//
+// Do is especially helpful in encoding nested attributes, attribute arrays,
+// or encoding arbitrary types (such as C structures) which don't fit cleanly
+// into an unsigned integer value.
+func (ae *AttributeEncoder) Do(t uint16, fn func() ([]byte, error)) {
+	if ae.err != nil {
+		return
+	}
+
+	b, err := fn()
+	if err != nil {
+		ae.err = err
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Encode returns the encoded bytes representing the attributes.
+func (ae *AttributeEncoder) Encode() ([]byte, error) {
+	if ae.err != nil {
+		return nil, ae.err
+	}
+
+	return MarshalAttributes(ae.attrs)
+}
--- a/example_attributeencoder_test.go
+++ b/example_attributeencoder_test.go
@@ -1,0 +1,75 @@
+package netlink_test
+
+import (
+	"encoding/hex"
+	"fmt"
+	"log"
+
+	"github.com/mdlayher/netlink"
+)
+
+// nested is a nested structure within out.
+type nested struct {
+	A, B uint32
+}
+
+// out is an example structure we will use to pack netlink attributes.
+type out struct {
+	Number uint16
+	String string
+	Nested nested
+}
+
+// encode is an example function used to adapt the ae.Do method
+// to encode an arbitrary structure.
+func (n nested) encode() func() ([]byte, error) {
+	return func() ([]byte, error) {
+		// Create an internal, nested netlink.NewAttributeEncoder that
+		// operates on the nested set of attributes.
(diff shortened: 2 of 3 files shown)


attribute.go
```diff
@@ -296,3 +296,132 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// MarshalBinary returns the encoded bytes representing the
```
mdlayher: Nit: Encode feels like a natural method name for this type, even though it's a bit of a departure from the usual names.

example_test.go
```diff
@@ -180,6 +181,77 @@ Nested:
...
 
+func ExampleAttributeEncoder_encode() {
+	// Create a netlink.AttributeEncoder that encodes to the same message
+	// as that decoded by the netlink.AttributeDecoder example.
+	ae, err := netlink.NewAttributeEncoder()
+	if err != nil {
+		log.Fatalf("failed to create attribute encoder: %v", err)
+	}
+
+	// nested is a nestes structure within out.
```
mdlayher: *nested

example_test.go
```diff
@@ -180,6 +181,77 @@ Nested:
...
+		},
+	}
+
+	// encodeNested is an example function used to adapt the ae.Do method
+	// to encode an arbitrary structure.
+	//
+	// It is recommended to create functions such as this one at the package
+	// level, but for this example, it is necessary to create a closure to
+	// fully show the usage of netlink.AttributeEncoder.
+	encodeNested := func(n *nested) func() ([]byte, error) {
```
mdlayher: We can probably avoid the pointer here for simplicity in the example: it's only necessary for decoding.

example_test.go
```diff
@@ -180,6 +181,77 @@ Nested:
...
+	ae.Do(3, encodeNested(&o.Nested))
+
+	b, err := ae.MarshalBinary()
+	// Any errors encountered during encoding (including any errors from
+	// encoding nested attributes) will be returned here.
+	if err != nil {
+		log.Fatalf("failed to encode attributes: %v", err)
+	}
+
+	fmt.Printf("Encoded netlink message follows:\n%s", hex.Dump(b))
```
mdlayher: It'd be nice to have an `// Output:` as in previous examples because then the package runs this as a test and verifies that the output is exactly what we expect.

attribute.go
```diff
@@ -296,3 +296,132 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	//
+	// If not set, the native byte order will be used.
+	ByteOrder binary.ByteOrder
+
+	attrs []Attribute
+	err   error
+}
+
+// NewAttributeEncoder creates an AttributeEncoder that encodes Attributes.
+func NewAttributeEncoder() (*AttributeEncoder, error) {
```
mdlayher: Can probably remove the error return here.  Then you could arguably remove the entire constructor.

There is the Go proverb of "make the zero value useful" which could apply well here:

```go
var ae netlink.AttributeEncoder
```

But then we also have to lazily set byte order on first call if that's the case. What do you think?

attribute.go
```diff
@@ -296,3 +296,132 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
 }
+
+// An AttributeEncoder provides a safe way to encode attributes.
+//
+// It is recommended to use an AttributeEncoder where possible instead of
+// calling MarshalBinary and using package nlenc directly.
+//
+// Errors from intermediate encoding steps are returned in the call to
+// MarshalBinary.
+type AttributeEncoder struct {
```
mdlayher: Do you think it makes sense to expose `func (ae *AttributeEncoder) Attrs() ([]netlink.Attribute, error)` to get the attribute stack out, in addition to or instead of returning bytes directly?

stapelberg:
Let me know if you want me to give this a go over in the google/nftables package once you’ve reviewed it.

mdlayher:
I made a couple of changes to this patch and decided to give it a go in wireguardctrl:

https://github.com/mdlayher/wireguardctrl/commit/4a8c868fa903db7607658ed35214115dc1104cb2

So far so good!  It all seems to come together nicely. I made a few tweaks that I think aid in the ergonomics of the type, and did a few doc updates:

https://gist.github.com/mdlayher/e911bfca231e4c8c339263b480aa6f6e

Curious to hear what you both think.  It seems useful to me to be able to easily embed a byte slice in attributes (common for things like IPs), and I also had a use case where I wanted the Go slice directly instead of the encoded attributes.

attribute.go
terinjokes: I originally had it as Encode, but changed it to match the interface. I'm happy to change it.

example_test.go
terinjokes: Would you be against making this a file example, so I can make this a package level function?

attribute.go
terinjokes: I would normally agree with you. However, here I prefer the symmetry with AttributeDecoder.

attribute.go
terinjokes: I have no preference. I have no need for this, so didn't include it.

terinjokes:
@mdlayher What I had at community day also had a `Bytes` type. I was hesitant in including here, since it`s a trivial adapter for ae.Do, and there`s a naming conflict with `nlenc.Bytes`

example_test.go
mdlayher: I had totally forgotten about whole file examples; seems reasonable to me to give it a go.

attribute.go
mdlayher: Yeah, I think Encode is good.  I'm considering removing the other exported MarshalBinary methods and such anyway.  No real need to conform to the stdlib interface.

attribute.go
mdlayher: That's fair.  I was thinking about doing something clever like memoizing byte order on the first call but it makes the internal code less clean.  How about we just remove the error return?

mdlayher:
Since this code lives in `netlink` and not `nlenc`, it ought to be just fine.  I agree that you can do it easily with a small adapter, but it seems reasonable to me to offer the method directly, because otherwise I'll probably write the same helper in my code repeatedly for use with Do.  It's nice and concise with a Bytes method, and general enough to make sense for many callers IMO.

example_attributeencoder_test.go
```diff
@@ -0,0 +1,75 @@
...
+// out is an example structure we will use to pack netlink attributes.
+type out struct {
+	Number uint16
+	String string
+	Nested nested
+}
+
+// encodeNested is an example function used to adapt the ae.Do method
+// to encode an arbitrary structure.
+func (n nested) enocdeNested() func() ([]byte, error) {
```
terinjokes: I'm not sold on the naming here, or if it should be a function on the type, or standalone.

attribute.go
terinjokes: I still haven't included this. Is there a specific case somewhere that needs this?

terinjokes:
> It's nice and concise with a Bytes method, and general enough to make sense for many callers IMO.

I'd argue the same for AttributesDecoder, as many of my `ad.Do` calls are to a function to copy bytes. I can open an issue.

attribute.go
```diff
@@ -296,3 +296,142 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+}
+
+// NewAttributeEncoder creates an AttributeEncoder that encodes Attributes.
+func NewAttributeEncoder() *AttributeEncoder {
+	return &AttributeEncoder{
+		ByteOrder: nlenc.NativeEndian(),
+	}
+}
+
+// Uint8 encodes the uint8 data for the specifiedeencoded.
```
mdlayher: *specified field

attribute.go
```diff
@@ -296,3 +296,142 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	ae.ByteOrder.PutUint64(b, v)
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// String encodes a null-terminated string for the specified field.
+func (ae *AttributeEncoder) String(t uint16, v string) {
```
mdlayher: Super nit: I like `s` for strings in a small scope.

attribute.go
```diff
@@ -296,3 +296,142 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	b := make([]byte, 8)
+	ae.ByteOrder.PutUint64(b, v)
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// String encodes a null-terminated string for the specified field.
```
mdlayher: Maybe `// String encodes s as a null-terminated string for the specified field.` to avoid any confusion around a possible null-terminated Go string.

attribute.go
```diff
@@ -296,3 +296,142 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Do is a general purpose function to encode arbitrary data into a field.
+//
+// Do is especially helpful in encoding nested attributes, attribute arrays,
+// or encoding arbitrary types which don't fit cleanly into an unsigned integer
```
mdlayher: I liked the "arbitrary types (such as C structures)" disclaimer I had on the decoder too, since it's a fairly common use case for netlink. Mind adding that statement?

attribute.go
```diff
@@ -296,3 +296,142 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{
+		Type: t,
+		Data: b,
+	})
+}
+
+// Encode returns the encoded bytes representing the
+// attributes.
```
mdlayher: Super nit: line 429 isn't too long so IMO don't line break before this.

example_attributeencoder_test.go
```diff
@@ -0,0 +1,75 @@
...
+// out is an example structure we will use to pack netlink attributes.
+type out struct {
+	Number uint16
+	String string
+	Nested nested
+}
+
+// encodeNested is an example function used to adapt the ae.Do method
+// to encode an arbitrary structure.
+func (n nested) encodeNested() func() ([]byte, error) {
```
mdlayher: Just "encode" should be succinct enough IMO.

mdlayher:
Yeah the bytes thing is tricky for decoding.  I suppose I'm not entirely against making a function that makes a copy of the bytes and returns it to the caller on the decoder.  Please do file an issue if you agree.

mdlayher:
LGTM thanks!

elliotmr:
Absolute first impression on using the library: could we use "type" and "value" as the argument names?

The usage is clear inside the function, but for documentation and auto-completion it would be better.

PR #96, opened 2018-09-05
conn_linux - stop syscall worker gracefully

ti-mo:
@mdlayher This is my take on  #92, key points being executing `wg.Wait()` before the syscall worker's input channel is closed, and using a done channel to gracefully terminate the goroutine.

This also reduces the amount of touching points with the mutex, making it a bit easier to modify. 
`Close()` is now executed off-thread because there is no longer any risk of misrouting messages to the wrong socket, and because slapping a write lock around an invocation of `do()` will attempt a recursive lock.

This change set turned out a bit bigger than intended, but removes `defer` in the hot path. I'm not sold on `doErr`, it feels verbose, but can't think of another approach either.

Will make a follow-up for unblocking pending `Receive()`.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -300,8 +300,10 @@ type sysSocket struct {
 	wg    *sync.WaitGroup
 	funcC chan<- func()
 
-	mu   sync.RWMutex
-	done bool
+	mu sync.RWMutex
+
+	done  bool
+	doneC chan<- bool
 }
 
 // newSysSocket creates a sysSocket that optionally locks its internal goroutine
@@ -315,6 +317,8 @@ func newSysSocket(lockThread bool) *sysSocket {
 	// Gophers Slack for providing this useful link.
 
 	funcC := make(chan func())
+	doneC := make(chan bool)
+
 	go func() {
 		// It is important to lock this goroutine to its OS thread for the duration
 		// of the netlink socket being used, or else the kernel may end up routing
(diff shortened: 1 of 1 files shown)


conn_linux.go
```diff
@@ -342,46 +346,64 @@ func newSysSocket(lockThread bool) *sysSocket {
 
 		defer wg.Done()
 
-		for f := range funcC {
-			f()
+		var f func()
```
mdlayher: IMO just `:=` on line 355 and remove.

conn_linux.go
```diff
@@ -342,46 +346,64 @@ func newSysSocket(lockThread bool) *sysSocket {
...
-func (s *sysSocket) do(f func()) {
+func (s *sysSocket) do(f func()) error {
 	done := make(chan bool, 1)
+
+	// All operations handled by this function are assumed to only
+	// read from s.done.
+	s.mu.RLock()
+
+	if s.done {
+		return syscall.EBADF
```
mdlayher: Need to unlock the mutex on this return path because you aren't using defer.

conn_linux.go
```diff
@@ -391,111 +413,97 @@ func (s *sysSocket) Socket(family int) error {
...
+	if doErr != nil {
+		return doErr
+	}
 
 	return err
 }
 
 func (s *sysSocket) Close() error {
+
+	var err error
```
mdlayher: Remove and `:=` on line 440.

conn_linux.go
```diff
@@ -391,111 +413,97 @@ func (s *sysSocket) Socket(family int) error {
...
-	// handling goroutine by closing the channel it is listening on,
-	// and lock out any future requests to avoid send on closed channel panics.
+	// Close the socket from the main thread, this operation has no risk
+	// of routing data to the wrong socket.
+	err = unix.Close(s.fd)
 	s.done = true
-	close(s.funcC)
+
+	// Signal the syscall worker to exit, wait for the WaitGroup to join,
+	// and close the job channel only when the worker is guaranteed to have stopped
```
mdlayher: Add a period.

mdlayher:
> This change set turned out a bit bigger than intended, but removes defer in the hot path. I'm not sold on doErr, it feels verbose, but can't think of another approach either.

Yep, not entirely sure it's avoidable, but it's a good start for sure.

> Will make a follow-up for unblocking pending Receive().

SGTM thanks.  This could be difficult and we may to do epoll or similar directly though AFAIK.

conn_linux.go
ti-mo: Doesn't this avoid an allocation?

mdlayher:
LGTM

PR #98, opened 2018-09-06
Set socket non-blocking

florianl:

This PR makes the socket non-blocking and will try receiving again, if error is EAGAIN.
So, you can close a socket, even if you try to receive something from it but nothing is there at the moment.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -57,6 +57,10 @@ func dial(family int, config *Config) (*conn, uint32, error) {
 		return nil, 0, err
 	}
 
+	if err := syscall.SetNonblock(sock.fd, true); err != nil {
+		return nil, 0, err
+	}
+
 	return bind(sock, config)
 }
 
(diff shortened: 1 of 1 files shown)


mdlayher:
Unfortunately, it seems that this will cause a syscall busy loop that will spin as fast as the CPU will let it go.

I think to do this efficiently, we'll need to either:

- integrate with the Go runtime network poller and use os.File.Read/Write directly (may not be doable)
- implement epoll or similar ourselves here

ti-mo:
Yup, also looked into epoll a bit:

- https://stackoverflow.com/questions/29237539/how-to-wake-up-a-pthread-which-is-sleeping-in-recvmsg
- https://gist.github.com/tevino/3a4f4ec4ea9d0ca66d4f
- https://github.com/mailru/easygo/blob/master/netpoll/epoll.go

Looks like a good solution, but will require significant changes to the core. Not sure how you would unit test such a contraption. :)

Implementing ReadWriteCloser seems possible, but I'm wondering how we will recognize message boundaries, since that effectively turns the socket into a stream. This is possible with some serial protocols that have markers to indicate message begin/end (`Scan()`), but I'm not sure if this is the case with netlink. Also, performance is not so great. Read timeouts and canceling is trivial to implement, there, though.

florianl:
Thanks for the feedback - Therefore, I will close this PR.

PR #102, opened 2018-09-07
netlink: exported API documentation cleanup

mdlayher:
Mostly going for consistency with other parts of the package; single letter names aren't generally as friendly.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -77,7 +77,7 @@ func Dial(family int, config *Config) (*Conn, error) {
 //
 // NewConn is primarily useful for tests. Most applications should use
 // Dial instead.
-func NewConn(c Socket, pid uint32) *Conn {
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -149,7 +149,8 @@ func (c *Conn) fixMsg(m *Message, ml int) {
 }
 
 // SendMessages sends multiple Messages to netlink. The handling of
-// m.Header.Length, Sequence and PID is the same as when calling Send.
+// the a Header's Length, Sequence and PID fields is the same as when
```
ti-mo: > the a

terinjokes:
In chat, I had suggested `attr` instead of the (IMHO, of course) more awkward `typ`. Otherwise, Look(ed) Good To Me.

PR #105, opened 2018-09-07
feat: add Bytes func to AttributeDecoder

terinjokes:
For symmetry with AttributeEncoder, adds a Bytes function to
AttriuteDecoder that copies the bytes to a new slice to avoid retaining
the original.


mdlayher:
LGTM!

PR #110, opened 2018-09-09
Network namespace support for Conn

ti-mo:
Bit of a work in progress, but I've been using this over the weekend to build a test suite around, and it works great (so far). There's some (well-founded) FUD going around about netns+goroutines, but this doesn't apply here.

We lock a 'leaf' goroutine with a narrow scope to an OS thread, there is no code path that can create goroutines from the locked thread. (these will end up in the 'main' namespace)

netns_linux.go is adapted from https://github.com/vishvananda/netns, should I attribute? I use this library to set up namespaces in my tests, too.

Let me know what you think, and let's discuss a way to integration test this.


mdlayher:
> netns_linux.go is adapted from https://github.com/vishvananda/netns, should I attribute?

Haven't looked at this yet, but this strikes me as potentially bad news.  This package is MIT but that one is Apache 2.0.  I am not a lawyer and cannot speak to the compatibility of the two, but I would really prefer an independent implementation of the primitives we need to avoid any potential legal ramifications.

ti-mo:
@mdlayher Re-written without `unix`.

conn.go
```diff
@@ -533,13 +533,6 @@ type Config struct {
...
-	// Experimental: do not lock the internal system call handling goroutine
-	// to its OS thread.  This may result in a speed-up of system call handling,
-	// but may cause unexpected behavior when sending and receiving a large number
-	// of messages.
-	//
-	// This should almost certainly be set to false, but if you come up with a
-	// valid reason for using this, please file an issue at
-	// https://github.com/mdlayher/netlink to discuss your thoughts.
-	NoLockThread bool
+	// Network namespace the Conn needs to operate in.
```
mdlayher: Add a clarification about zero-value:

```go
// Network namespace the Conn needs to operate in.  If set to 0,
// no network namespace will be applied.
NetNS int
```

conn_linux.go
```diff
@@ -50,10 +50,12 @@ func dial(family int, config *Config) (*conn, uint32, error) {
...
 
-	lockThread := !config.NoLockThread
-	sock := newSysSocket(lockThread)
+	sock, err := newSysSocket(config)
+	if err != nil {
+		return nil, 0, err
+	}
 
-	if err := sock.Socket(family); err != nil {
+	if err = sock.Socket(family); err != nil {
```
mdlayher: Nit: I like `:=` for these small scopes so this `err` is entirely contained.

conn_linux.go
```diff
@@ -344,32 +346,50 @@ func newSysSocket(lockThread bool) *sysSocket {
...
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
+
+		// The user requested the Conn to operate in a non-default network namespace.
+		if config.NetNS != 0 {
+
+			// Get the current namespace of the thread the goroutine is locked to.
+			origNetNS, err := getThreadNetNS()
+			if err != nil {
+				errC <- err
```
mdlayher: On error need to return.

Should probably move defer on line 393 up to roughly line 368.

conn_linux.go
```diff
@@ -344,32 +346,50 @@ func newSysSocket(lockThread bool) *sysSocket {
...
+			}
+
+			// Set the network namespace of the current thread using
+			// the file descriptor provided by the user.
+			err = setThreadNetNS(config.NetNS)
+			if err != nil {
+				errC <- err
```
mdlayher: On error need to return.

netns_linux.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+// getThreadNetNS gets the network namespace file descriptor of the thread the current goroutine
+// is running on. Make sure to call runtime.LockOSThread() before this so the goroutine does not
+// get scheduled on another thread in the meantime.
+func getThreadNetNS() (int, error) {
```
mdlayher: I think we may want to return `(*os.File, error)` so the file handle can be closed appropriately when no longer needed.

ti-mo: Good idea, but namespace handles work a bit differently. The namespace handle we obtain here is not actually created in userspace; its lifecycle is managed by the kernel (created with a syscall). Namespace references in userspace do not count towards the 'refcount', only the amount of threads in the namespace do.

A namespace ceases to exist when the last thread in it exits, therefore I don't believe it makes sense to wrap it into an `os.File`, because the lifecycle of the namespace (and its handle) are not attached to the lifecycle of the `os.File` object at all.

PR #116, opened 2018-11-23
Align nl message length

florianl:

While working on florianl/go-nfqueue, I've noticed netlink messages from the kernel without correct alignment. Besides the alignment, these messages are fine and can be parsed without problem.

I've traced down this issue to `func (c *conn) Receive() ([]Message, error)` in `conn_linux.go`, where `syscall.ParseNetlinkMessage()` failed with `EINVAL`.
E.g. `n` for `Recvmsg(b, nil, unix.MSG_PEEK)` is 171, `n` for `Recvmsg(b, nil, 0)` is also 171 and `syscall.ParseNetlinkMessage(b[:n])` failed because `nlmAlignOf()` in `syscall/netlink_linux.go` returns 172.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -129,6 +129,12 @@ func (c *conn) Send(m Message) error {
 	return c.s.Sendmsg(b, nil, addr, 0)
 }
 
+// Round the length of a netlink message up to align it properly.
+// copy from syscall/netlink_linux.go:nlmAlignOf()
+func nlmAlignOf(msglen int) int {
+	return (msglen + syscall.NLMSG_ALIGNTO - 1) & ^(syscall.NLMSG_ALIGNTO - 1)
+}
+
 // Receive receives one or more Messages from netlink.
 func (c *conn) Receive() ([]Message, error) {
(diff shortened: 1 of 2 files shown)


terinjokes:
This EINVAL behavior was implemented in [CL26990](https://go-review.googlesource.com/c/go/+/26990/). The change is the same as elsewhere, specifically https://github.com/subgraph/go-nfnetlink/commit/756bdc512690c028c3d4aa1c2b2bd514524e9396.

LGTM.

mdlayher:
At a glance this looks reasonable to me!  Can you please add a test for the specific case you're describing to ensure the alignment logic works as expected?

conn_linux.go
```diff
@@ -165,6 +171,8 @@ func (c *conn) Receive() ([]Message, error) {
 		return nil, errInvalidFamily
 	}
 
+	n = nlmAlignOf(n)
```
mdlayher: Can just use nlmsgAlign defined in align.go and remove the new code.

florianl: sorry that I missed this existing function before.

PR #124, opened 2019-02-21
netlink: enable usage of the network poller

mdlayher:
WORK IN PROGRESS

Factor out goroutine management logic from sysSocket. Add a new
lockedNetNSGoroutine type, representing a goroutine
which executes code on a specific operating system thread.
As its name suggests, a lockedNetNSGoroutine can be configured to
operate in a non-default network namespace.

Add read, write, and control methods to sysSocket, which add more
context and semantics at previous call sites of (*sysSocket).do.

Switch sysSocket to use an *os.File internally, instead of an
integer file descriptor. Use the poller to make calls against
the socket. Add Set{Deadline,ReadDeadline,WriteDeadline} methods to
Conn.

Add tests to verify that read deadlines work.

---

Hey @acln0, now that draft PRs are a thing, I'm opening this up so more folks can see it! In addition, I'll have a commit for you to review shortly that fixes the merge problem I just created with #121.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -25,6 +25,7 @@ var (
 	errReadWriteCloserNotSupported = errors.New("raw read/write/closer not supported")
 	errMulticastGroupsNotSupported = errors.New("multicast groups not supported")
 	errBPFFiltersNotSupported      = errors.New("BPF filters not supported")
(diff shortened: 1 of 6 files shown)


mdlayher:
Superseded by #125.

PR #125, opened 2019-02-21
netlink: enable usage of the network poller

mdlayher:
Apparently I can't do a PR from my tree into @acln0's, so here's a new pull request that supersedes #124.

The comments from the original commit:

---

WORK IN PROGRESS

Factor out goroutine management logic from sysSocket. Add a new
lockedNetNSGoroutine type, representing a goroutine
which executes code on a specific operating system thread.
As its name suggests, a lockedNetNSGoroutine can be configured to
operate in a non-default network namespace.

Add read, write, and control methods to sysSocket, which add more
context and semantics at previous call sites of (*sysSocket).do.

Switch sysSocket to use an *os.File internally, instead of an
integer file descriptor. Use the poller to make calls against
the socket. Add Set{Deadline,ReadDeadline,WriteDeadline} methods to
Conn.

Add tests to verify that read deadlines work.

---

I have applied a patch of mine own to fix the merge conflicts caused by my setsockopt changes.

Marking this as draft again.


acln0:
Thanks for opening this.

The PR still needs some work before it's ready to review. I need to take a good look at everything again, in the near future. I'll also send a patch to change the tag from `tip` to `1.12`, which is what it should be.

Perhaps there should be travis builds for both `1.11` and `1.12`, to ensure that the old behavior is kept intact, while all the new stuff for `1.12` works correctly. In the meantime, I wonder what this last `staticcheck` failure is about: https://travis-ci.org/mdlayher/netlink/jobs/496784902#L118. ISTM that `fdcall_lt_1.12.go` is the code the build is actually using, since it's not building with `-tags=tip`.

Given that you have already committed on top of my branch, I assume you want me to base my future commits on this branch, is that correct? Will you be merging the whole branch without squashing?

mdlayher:
> Perhaps there should be travis builds for both 1.11 and 1.12, to ensure that the old behavior is kept intact, while all the new stuff for 1.12 works correctly. 

Yes please, that'd be great.

> In the meantime, I wonder what this last staticcheck failure is about: https://travis-ci.org/mdlayher/netlink/jobs/496784902#L118. 

That's a new one to me. I imagine if we set up 1.11 and 1.12 jobs we can catch both issues, bur if we end up needing temporary override directives, I'm also okay with that.

> Given that you have already committed on top of my branch, I assume you want me to base my future commits on this branch, is that correct? Will you be merging the whole branch without squashing?

I was going to try to send a PR to you, but it appears I can't, so I ended up creating my own branch on top of yours. If you'd like to either send more PRs against my version of the branch, that'd work, or otherwise feel free to cherry pick my merge conflict resolution commit and keep working on your end.

I wasn't sure what the best way to address the merge conflict was, so I attempted to resolve it myself. I suspect the changes here will be complex enough that I'm okay with not squashing and just merging the branch later.

mdlayher:
Go 1.12 is out! @acln0, would you like to continue driving this forward or would you like me to pitch in?  Not totally sure what to do with all of the tests as of yet.

acln0:
I want to drive this forward, and I'm going to work on it starting tomorrow. It's been a busy week for me. I have some changes to the tests in mind, so let's talk about that once I send some new commits.

Could you please grant me push access, as we discussed? I think it would be easiest if I could push commits right to this branch.

mdlayher:
Done! I will make sure to not force push to this branch.

acln0:
Thanks. I'll be in touch soon.

fdcall_gteq_1.12.go
```diff
@@ -0,0 +1,48 @@
+//+build tip
```
florianl: Can we use the "go1.12" build tag?

acln0: We do now.

acln0:
Okay, I think have made the changes I had in mind for the PR to be in a state where it can be tested and reviewed. The CI runs use 1.10+, since that's when `*os.File` gained its deadline methods. As of 1.12 being released, 1.10 is not officially supported anymore, but I thought it was a good cut-off point for now, since 1.12 is very fresh right now.

I made quite a few commits, but I can squash them down if necessary.

Please take a look.

rawconn_lt_1.12.go
```diff
@@ -0,0 +1,29 @@
...
+type rawConn struct {
+	fd uintptr
+}
+
+func (rc *rawConn) Control(f func(fd uintptr)) error {
+	f(rc.fd)
+	return nil
+}
+
+// TODO(mdlayher): implement Read and Write?
```
mdlayher: Super nit: but future note for me just to drop this TODO; it's better to just not support it on <1.12.

conn_linux.go
```diff
@@ -450,14 +439,17 @@ func (s *sysSocket) Socket(family int) error {
 		return err
 	}
 
-	s.fd = fd
+	if err := setBlockingMode(fd); err != nil {
+		return err
+	}
+	s.fd = os.NewFile(uintptr(fd), "netlink")
```
mdlayher: I would like to see a comment here noting that passing a non-blocking FD to this function registers it with the poller, and a link to https://golang.org/pkg/os/#NewFile, for future readers.

fdcall_lt_1.12.go
```diff
@@ -0,0 +1,24 @@
+//+build !go1.12,linux
+
+package netlink
+
+import "os"
+
+func setBlockingMode(sysfd int) error {
```
mdlayher: I think it'd be nice to add a comment in this file that it exists for compatibility with older versions of Go, although deadlines will be non-functional.

conn.go
```diff
@@ -394,6 +393,44 @@ func (c *Conn) RemoveBPF() error {
...
 
+// A deadlineSetter is a Socket that supports setting deadlines.
+type deadlineSetter interface {
+	Socket
+	SetDeadline(time.Time) error
+	SetReadDeadline(time.Time) error
+	SetWriteDeadline(time.Time) error
+}
+
+// SetDeadline sets the read and write deadlines associated with the connection.
```
mdlayher: I think it's worth adding a note to each of these functions about their Go version compatibility:

```go
// SetDeadline sets the read and write deadlines associated with the connection.
//
// Deadline functionality is only supported on Go 1.12+. Calling this function on older versions
// of Go will result in an error.
```

mdlayher:
LGTM!

PR #128, opened 2019-02-23
netlink: introduce OpError type, use throughout

### LINKED ISSUE #126: netlink: consider addition of a proper Error type
@acln0 left this excellent comment on #123:

> Throughout the standard library, errors from system calls are decorated with additional details, e.g. `os.SyscallError`, `net.OpError`. Package netlink doesn't do any of this at the moment, as far as I can tell. It seems to return `syscall.Errno` values up to the callers.
>
> Perhaps this is something worth thinking about. For example, when does `sendmsg(2)` fail on a netlink socket? If and when it does, is the `syscall.Errno` value enough for the caller to be able to diagnose the cause of the error?
>
> This circles back to the `netlink.Error` naming issue I mentioned above. Is there any point in having a type that captures operational errors and decorates them with additional context? If there is, what would the type be named? Would it be `netlink.Error`? Or perhaps `netlink.OpError`?
>
> Personally, all the errors I've gotten while using netlink have been in the form of netlink error messages, rather than system call errors from operations such as calling `conn.Send`, so I haven't had this problem. That being said, I have not used netlink in very advanced ways, so perhaps I have missed something. But having noticed this detail about error handling while I was working on the code, I thought I would at least bring it up for discussion.

Thus far I've been pretty lax about which errors this package should expose. I believe we could do a lot better with the addition of a proper Error type, which not only exposes the underlying system call error, but perhaps some trace points for better diagnosis of problems than a plain `EINVAL` or similar.  It'd also be worth considering the error values proposal if that work ends up landing in Go 1.13, so that we can support nicely formatted, wrap/unwrap-able errors.

mdlayher:
Fixes #126.

/cc @acln0

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -252,7 +252,14 @@ func (c *Conn) Receive() ([]Message, error) {
 
 // receive is the internal implementation of Conn.Receive, which can be called
 // recursively to handle multi-part messages.
-func (c *Conn) receive() ([]Message, *OpError) {
+func (c *Conn) receive() ([]Message, error) {
+	// NB: All non-nil errors returned from this function *must* be of type
+	// OpError in order to maintain the appropriate contract with callers of
+	// this package.
+	//
+	// This contract also applies to functions called within this function,
+	// such as checkMessage.
+
 	var res []Message
 	for {
 		msgs, err := c.sock.Receive()
--- a/conn.go
+++ b/conn.go
@@ -12,25 +12,6 @@ import (
 	"golang.org/x/net/bpf"
 )
 
-// Error messages which can be returned by Validate.
-var (
-	errMismatchedSequence = errors.New("mismatched sequence in netlink reply")
-	errMismatchedPID      = errors.New("mismatched PID in netlink reply")
-	errShortErrorMessage  = errors.New("not enough data for netlink error code")
-)
-
-// Errors which can be returned by a Socket that does not implement
-// all exposed methods of Conn.
-var (
-	errReadWriteCloserNotSupported = errors.New("raw read/write/closer not supported")
-	errMulticastGroupsNotSupported = errors.New("multicast groups not supported")
-	errBPFFiltersNotSupported      = errors.New("BPF filters not supported")
-	errDeadlinesNotSupported       = errors.New("deadlines not supported")
-	errOptionsNotSupported         = errors.New("options not supported")
-	errSetBufferNotSupported       = errors.New("setting buffer sizes not supported")
-	errSyscallConnNotSupported     = errors.New("syscall.RawConn operation not supported")
-)
-
 // A Conn is a connection to netlink.  A Conn can be used to send and
 // receives messages to and from netlink.
 //
@@ -110,7 +91,14 @@ func (c *Conn) debug(fn func(d *debugger)) {
 
 // Close closes the connection.
 func (c *Conn) Close() error {
-	return c.sock.Close()
+	if err := c.sock.Close(); err != nil {
+		return &OpError{
+			Op:  "close",
+			Err: err,
+		}
+	}
+
+	return nil
 }
 
 // Execute sends a single Message to netlink using Conn.Send, receives one or more
@@ -177,7 +165,10 @@ func (c *Conn) SendMessages(messages []Message) ([]Message, error) {
 			d.debugf(1, "send msgs: err: %v", err)
 		})
 
-		return nil, err
+		return nil, &OpError{
+			Op:  "send-messages",
+			Err: err,
+		}
 	}
 
 	return messages, nil
@@ -215,7 +206,10 @@ func (c *Conn) Send(message Message) (Message, error) {
 			d.debugf(1, "send: err: %v", err)
 		})
 
-		return Message{}, err
+		return Message{}, &OpError{
+			Op:  "send",
+			Err: err,
+		}
 	}
 
 	return message, nil
@@ -258,12 +252,15 @@ func (c *Conn) Receive() ([]Message, error) {
 
 // receive is the internal implementation of Conn.Receive, which can be called
 // recursively to handle multi-part messages.
-func (c *Conn) receive() ([]Message, error) {
+func (c *Conn) receive() ([]Message, *OpError) {
 	var res []Message
 	for {
 		msgs, err := c.sock.Receive()
 		if err != nil {
-			return nil, err
+			return nil, &OpError{
+				Op:  "receive",
+				Err: err,
+			}
 		}
 
 		// If this message is multi-part, we will need to perform an recursive call
@@ -328,13 +325,13 @@ func (rwc *fileReadWriteCloser) Close() error { return rwc.f.Close() }
 // performed using Conn and the raw io.ReadWriteCloser do not conflict with
 // each other.  In almost all scenarios, only one of the two should be used.
 func (c *Conn) ReadWriteCloser() (io.ReadWriteCloser, error) {
-	fc, ok := c.sock.(filer)
+	conn, ok := c.sock.(filer)
 	if !ok {
-		return nil, errReadWriteCloserNotSupported
+		return nil, notSupported("read-write-closer")
 	}
 
 	return &fileReadWriteCloser{
-		f: fc.File(),
+		f: conn.File(),
 	}, nil
 }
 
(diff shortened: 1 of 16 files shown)


errors.go
```diff
@@ -0,0 +1,68 @@
...
+		return os.IsNotExist(err.Err)
+	default:
+		return os.IsNotExist(err)
+	}
+}
+
+var _ error = &OpError{}
+
+// An OpError is an error produced as the result of a failed netlink operation.
+type OpError struct {
```
mdlayher: Once #124 lands, I'd like to have this type implement `net.Error` as well so it's easy to check for timeouts and temporary/permanent errors.

conn.go
```diff
@@ -257,12 +252,15 @@ func (c *Conn) Receive() ([]Message, error) {
 
 // receive is the internal implementation of Conn.Receive, which can be called
 // recursively to handle multi-part messages.
-func (c *Conn) receive() ([]Message, error) {
+func (c *Conn) receive() ([]Message, *OpError) {
```
acln0: Returning `*OpError` concretely seems dangerous in view of the possible "nil but not nil interface" mistake, unless I'm missing something about how this code works.

acln0: Returning the concrete `*OpError` seems dangerous in view of the possible "nil but not nil interface" mistake.

To expedite the conversation, I'm quoting your response from Slack:

> You raise a valid point; my thought here was to ensure that anything that comes out of this function has to be an OpError for consistency. I could potentially do something at the call site to check, or I could change this back and leave a note that all errors must be of type OpError or nil.

I like the latter option, but no strong opinions either way.

conn.go
```diff
@@ -176,7 +165,10 @@ func (c *Conn) SendMessages(messages []Message) ([]Message, error) {
 			d.debugf(1, "send msgs: err: %v", err)
 		})
 
-		return nil, err
+		return nil, &OpError{
+			Op:  "send-messages",
+			Err: err,
```
acln0: I'm leaving this comment here for lack of a better place. The high level operation is `"send-messages"`, but we know nothing about `err` itself. Presumably, the implementation of `c.sock` has made a system call, and the system call has returned an error. For completeness, I think the system call (e.g. `sendmsg` ) should be recorded in the form of an `*os.SyscallError`, and this should happen inside the `Socket` implementation.

The code right here should not change, but I think all the code that is making system calls should.

errors.go
acln0: Agreed. I think it should be reasonably easy to implement. `syscall.Errno` has a `Temporary() bool` and a `Timeout() bool`. We should unwrap `Err`, and if we have a `syscall.Errno` (or an `*os.SyscallError` wrapping a `syscall.Errno`), we should call those methods.

Something very similar to this: https://golang.org/src/net/net.go?s=15673:15707#L492

errors.go
```diff
@@ -0,0 +1,68 @@
...
+// should always use netlink.IsNotExist, rather than os.IsNotExist, when
+// checking for specific netlink-related errors.
+//
+// Errors types created by this package, such as OpError, can be used with
+// IsNotExist, but this function also defers to the behavior of os.IsNotExist
+// for unrecognized error types.
+func IsNotExist(err error) bool {
+	switch err := err.(type) {
+	case *OpError:
+		// TODO(mdlayher): more error handling logic?
```
acln0: > more error handling logic?

Only if it's netlink-specific, I imagine.

I will note that `os.IsNotExist` is good enough even if we end up using `*os.SyscallError` for the inner errors, since the `os` package deals with the unwrapping.

conn.go
mdlayher: SGTM!

errors.go
mdlayher: TIL! To clarify though, this function is still required in order to unpack OpError.Err properly since it's a struct field right?  I assumed it made more sense to have our own version with these sorts of internal unpacking behaviors.

acln0: Yes, that's right right. My comment about `os.IsNotExist` refers strictly to the inner error (the error in the `Err` field of `*OpError`).

mdlayher:
Because #125 will introduce some churn, I'm going to merge that one first, then rectify the changes there with this PR.

conn.go
mdlayher: This has been applied, review when ready!

message.go
```diff
@@ -247,13 +251,19 @@ func checkMessage(m Message) error {
...
+			Op:  "receive",
+			Err: errShortErrorMessage,
+		}
 	}
 
 	if c := nlenc.Int32(m.Data[0:4]); c != success {
 		// Error code is a negative integer, convert it into
 		// an OS-specific system call error
-		return newError(-1 * int(c))
+		return &OpError{
```
mdlayher: I'm having a difficult time deciding if this change is needed, and whether or not the internal newError should do os.NewSyscallError.

These are errors derived from netlink messages, not from the system calls used to perform them.  What's the correct answe here?  Should we have another type like MessageError or something to differentiate?

acln0: ISTM like the system call itself was successful, but the action the message was meant to take was not. Therefore, it feels like this is not the right place for a `SyscallError`, since `SyscallError` records an error from a specific call.

Where would the hypothetical `MessageError` fit in? Would it be returned instead of `OpError` on some code paths? Or would it be returned in the `Err` field of an `OpError`?

`newError` returns a `syscall.Errno`, and there is not much more context to it than that. This time, the _caller_ has the context, because they know what high level operation they are doing. How about the following lightweight convention: if `.Err` is an `*os.SyscallError`, then the caller knows that it is a system call that failed; if `.Err` is a naked `syscall.Errno`, then the operation itself has failed. If the caller _really_ cares about this beyond what `IsNotExist` or similar functions can offer, then we can simply document this convention.

conn_linux.go
```diff
@@ -61,11 +58,11 @@ func dial(family int, config *Config) (*conn, uint32, error) {
 
 	sock, err := newSysSocket(config)
 	if err != nil {
-		return nil, 0, err
+		return nil, 0, os.NewSyscallError("netlink-sys-socket", err)
```
acln0: I don't think this is quite right. This call to `newSysSocket` only fails if this call fails:
https://github.com/mdlayher/netlink/blob/master/conn_linux.go#L597

That, in itself, can only fail here:
https://github.com/mdlayher/netlink/blob/master/netns_linux.go#L18

And here:
https://github.com/mdlayher/netlink/blob/master/netns_linux.go#L26

In `getThreadNetNS`, we already get a `SyscallError`, from `os.Open`, so there is nothing to do there.

In `setThreadNetNS`, we should wrap the error in `os.NewSyscallError("setns", ...)`.

Then, the error should just be bubbled up without further decoration from this site. This should be documented in case it changes in the future, and we _do_ end up needing more decoration; but that seems unlikely.

conn.go
acln0: Looks good. Marking as resolved.

errors.go
mdlayher: This is done in the newest commit.

conn_linux.go
mdlayher: Agreed, and will update. To clarify, where should this be documented?

acln0: Right here would be fine, I think.

mdlayher: Do you think I should do `_ = err.(*os.SyscallError)` to ensure the contract isn't violated?

mdlayher: It appears this won't work; note that os.Open says it returns a PathError: https://golang.org/pkg/os/#Open.

Thoughts?

acln0: That wouldn't be exactly right either, because https://github.com/mdlayher/netlink/blob/master/netns_linux.go#L18 is probably an `*os.PathError` with an inner `*os.SyscallError`.

mdlayher: Understood, I'll just document it and state that bare syscall errors should be wrapped.

message.go
mdlayher: I see what you mean!  I will revert this, but I will document OpError.Err to explain the convention.

acln0: I'm confused. Is there anything to revert here? The change seems fine as-is.

mdlayher: I think I got confused as well!  I'll proceed according to the convention you mentioned above.

mdlayher: Okay, I updated the docs and did a new commit with nltest changes too to check the behavior!

acln0: Excellent. This behavior LGTM now.

errors.go
```diff
@@ -0,0 +1,110 @@
...
+	// Most callers should inspect Err using a helper such as IsNotExist.
+	Err error
+}
+
+func (e *OpError) Error() string {
+	if e == nil {
+		return "<nil>"
+	}
+
+	return fmt.Sprintf("netlink %q: %v", e.Op, e.Err)
```
acln0: Tiny nit: the `net` package does not quote the operation string: https://golang.org/src/net/net.go#L455. To me, `netlink send: something went wrong` looks better than `netlink "send": something went wrong` since I'm used to seeing errors from package `net`.

conn_linux.go
```diff
@@ -161,24 +158,16 @@ func (c *conn) Receive() ([]Message, error) {
...
-		return nil, errInvalidFamily
+		return nil, os.NewSyscallError("recvmsg", err)
 	}
 
 	n = nlmsgAlign(n)
 
 	raw, err := syscall.ParseNetlinkMessage(b[:n])
 	if err != nil {
-		return nil, err
+		return nil, os.NewSyscallError("syscall-parse-netlink-message", err)
```
acln0: Yet another pedantic nit, sorry!

This code is calling a function from the `syscall` package, but that function is not actually making any system calls: https://golang.org/src/syscall/netlink_linux.go#L116.

It doesn't seem like this is a condition that the caller can recover from anyway. Perhaps a `fmt.Errorf("failed to parse netlink message: %v", err)` would be more semantically precise.

PR #131, opened 2019-04-03
netlink: serialize calls to Conn.Execute, lock in concurrency behaviors

mdlayher:
This enables and verifies:

- many concurrent calls to Send/Receive, relying on the kernel to deal with socket concurrency
- serialized calls to Execute, to verify that individual request/response pairs actually line up (the most common use case)
- Receive and other calls can always be unblocked by Close

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -236,7 +236,7 @@ func (c *Conn) Receive() ([]Message, error) {
 	return c.lockedReceive()
 }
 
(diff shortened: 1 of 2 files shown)


mdlayher:
The Travis failures appear to be related to Go 1.10 and staticcheck. No need for concern.

PR #133, opened 2019-04-12
netlink: append to Receive buffer instead of allocating a new one

mdlayher:
I'd like a bit of a sanity check here, but this is surely the better approach, right? If I understand, previously I was totally discarding the bytes of `b` and allocating a brand new slice every time, instead of appending to the existing one.


mdlayher:
> Presumably `len(b) == cap(b)`, so you need a new allocation either way.

Thanks @acln0! Closing.

PR #134, opened 2019-04-18
netlink: note Gophers Slack #networking channel

mdlayher:


README.md
```diff
@@ -6,6 +6,10 @@ MIT Licensed.
 For more information about how netlink works, check out my blog series
 on [Linux, Netlink, and Go](https://medium.com/@mdlayher/linux-netlink-and-go-part-1-netlink-4781aaeeaca8).
 
+If you have any questions or you'd like some guidance, please join us on
+[Gophers Slack](https://invite.slack.golangbridge.org) in the `#networking`
+channel!
```
bradbeam: Should you also mention #netlink ?

insomniacslk: it's very silent at the moment. Maybe we could redirect people there if the volume of netlink conversations increases too much for #networking ?

mdlayher: ^ Yep, that was basically my plan. I don't think we've hit critical mass yet.

PR #139, opened 2019-05-30
Add socket option NETLINK_EXT_ACK

florianl:

This PR adds support for the socket option NETLINK_EXT_ACK. 

https://lwn.net/Articles/719721/

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -428,6 +428,7 @@ const (
 	NoENOBUFS
 	ListenAllNSID
 	CapAcknowledge
+	ExtendAckReporting
 )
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -428,6 +428,7 @@ const (
 	NoENOBUFS
 	ListenAllNSID
 	CapAcknowledge
+	ExtendAckReporting
```
mdlayher: Let's go with "ExtendedAcknowledge" like we talked about on Slack, but otherwise this all LGTM.

PR #141, opened 2019-06-04
netlink: add CallingThreadNetNS to enter caller's network namespace

mdlayher:
This is a little bit tricky.

A user has created an application which locks OS thread and enters its own network namespace before using `wgctrl.New()` to manipulate WireGuard devices in that namespace. Because wgctrl is backed by this package but provides no explicit tunable at the moment to enter a namespace via the netlink connection, it would seem the correct behavior is to automatically inherit the calling thread's network namespace in the new goroutine.

I'm not an expert on namespaces, but I believe this change to be backward compatible. The majority of callers who are not specifying NetNS today expect to inherit the network namespace of the calling process. This should account for that use case, as well as that of a caller who has manipulated the calling goroutine's thread into another namespace as well.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -569,6 +569,11 @@ const (
 	// network namespaces. See the documentation of runtime.LockOSThread and
 	// runtime.UnlockOSThread for details.
 	//
+	// A Conn is capable of entering a specific network namespace on its own,
+	// without any OS thread locking by the caller, but this option is useful
+	// for cases where the caller has already locked the calling OS thread and
+	// manipulated the network namespace of the calling thread.
+	//
 	// This option is useful in very specific circumstances, and must be used
 	// with great care.
 	CallingThreadNetNS = -1
--- a/conn.go
+++ b/conn.go
@@ -18,6 +18,10 @@ import (
 // A Conn is safe for concurrent use, but to avoid contention in
 // high-throughput applications, the caller should almost certainly create a
 // pool of Conns and distribute them among workers.
+//
+// A Conn is capable of manipulating netlink subsystems from within a specific
+// Linux network namespace, but special care must be taken when doing so. See
+// the documentation of Config for details.
 type Conn struct {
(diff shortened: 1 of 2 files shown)


mdlayher:
Ugh, it would appear that setting the namespace is a privileged operation, and this package explicitly tries to avoid privileged operations where possible.

As an alternative, I thought of doing something like `&netlink.Config{NetNS: netlink.InheritNetNS}` and then setting that in wgctrl or similar, but then wgctrl also requires special privileges. Suggestions appreciated.

mdlayher:
Since wgctrl requires special privileges anyway, perhaps that approach is the correct way forward:

```
✔ ~/src/golang.zx2c4.com/wireguard/wgctrl/cmd/wgctrl [master|⚑ 1] 
09:25 $ ./wgctrl 
2019/06/04 09:25:03 failed to get devices: operation not permitted
✘-1 ~/src/golang.zx2c4.com/wireguard/wgctrl/cmd/wgctrl [master|⚑ 1] 
09:25 $ sudo ./wgctrl 
interface: wg0 (Linux kernel)
```

conn.go
```diff
@@ -554,13 +558,37 @@ func Validate(request Message, replies []Message) error {
...
+	// explicitly specified in order to do so.
+	//
+	// Note that the Go runtime multiplexes many goroutines onto many OS threads,
+	// and special care must be taken when manipulating thread state such as
+	// network namespaces. See the documentation of runtime.LockOSThread and
+	// runtime.UnlockOSThread for details.
+	//
+	// This option is useful in very specific circumstances, and must be used
+	// with great care.
+	CallingThreadNetNS = -1
```
mdlayher: Name is up for discussion, but this felt most explicit. As for the value, I figure any negative number will do fine.

mdlayher:
Updated to take a different approach that requires setting an explicit constant to make this work. This constant will be used in wgctrl.

PR #142, opened 2019-06-04
netlink: best effort switching to calling thread's network namespace

mdlayher:
An alternative to #141, for https://github.com/WireGuard/wgctrl-go/issues/58.

If no network namespace is set, make a best effort to switch to the calling thread's namespace anyway. If the caller has permission, it works. If they don't, nothing happens.

If the caller explicitly sets a namespace and they do not have permission, an error is returned.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -558,42 +558,27 @@ func Validate(request Message, replies []Message) error {
 	return nil
 }
 
-const (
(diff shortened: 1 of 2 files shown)


mdlayher:
Merging quickly to get rid of the constant I previously added. I'm working on tests for this too.

PR #145, opened 2019-06-14
netlink: support for systems with network namespaces disabled

### LINKED ISSUE #144: netlink: network namespaces can be disabled in kernel configuration
When I try to use the library in order to open a generic netlink socket I get the error message:
```
panic: failed to dial: open /proc/self/task/2281/ns/net: no such file or directory
```

It seems that this library expects a feature to be enabled in my kernel which is not. Do you happen to know which kernel build configuration/feature is responsible for this subdirectory? It might be handy to document this requirement as well.

mdlayher:
The `conn_linux.go` changes are required to make this testable, but this should do the trick.

Fixes #144.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -338,7 +338,8 @@ type sysSocket struct {
 // newSysSocket creates a sysSocket that optionally locks its internal goroutine
 // to a single thread.
 func newSysSocket(config *Config) (*sysSocket, error) {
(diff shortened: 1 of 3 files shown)


mdlayher:
Tests pass and got confirmation from @0x5a17ed that the issue is resolved, so merging.

Issue #146, opened 2019-08-25
Is locking of go-routine to thread mandatory?


dstiliadis:
@mdlayher first, I want to thank you for the great work you have done with this library.

While doing some performance tests, I identified some performance issues 
with the thread lock. 
https://github.com/mdlayher/netlink/blob/739ceab8cfd707a27eed3220823ec86ea547911a/conn_linux.go#L664

If you do a pprof and look at the flame graph, you will notice that this threads consume
a constant amount of CPU that is relatively high, and also create caching issues. 

My theory is that this is indeed needed when you have multiple clients that 
connect in different namespaces (or for that matter, even single client that connects
to a namespace other than the parent process namespace). But it should not be 
needed if all clients are in the same namespace as the parent. The issue with go
and namespaces is that the thread must be locked inside a namespace, but if all
clients are in the main process namespace there should be no need to lock the thread.

In tests that we did, we did not notice the problem if all clients are in the parent
namespace.

I was wondering if you have a specific test-case scenario where the bug was showing up.

If not, would it make sense to send a patch, where the thread lock is only activated for
any remote namespace call?

Thanks,

PR #147, opened 2019-08-31
Provide configuration option for not locking go-routines to threads

dstiliadis:
This is an implementation of the configuration option that will allow callers to 
define that the thread lock must be bypassed. 

See discussion here for details: https://github.com/mdlayher/netlink/issues/146

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -593,7 +593,6 @@ type Config struct {
 	// impact. In these cases, the caller can disable thread locking if performance
 	// considerations are of interest. If disabled, it is the responsibility of
 	// the caller to make sure that all threads are running in the correct namespace.
-	// When DisableNSLockThread is set, the caller cannot set the NetNS value. An error
-	// will be returned.
+	// When DisableNSLockThread is set, the caller cannot set the NetNS value.
 	DisableNSLockThread bool
 }
--- a/conn.go
+++ b/conn.go
@@ -584,16 +584,16 @@ type Config struct {
 	// to 0.
 	NetNS int
 
-	// DisableNSLockThread
+	// DisableNSLockThread disables package netlink's default goroutine thread locking behavior.
 	//
-	// By default, the library will lock the processing go routine to its corresponding
+	// By default, the library will lock the processing goroutine to its corresponding
 	// thread in order to enable communication over netlink to a different namespaces.
 	// In some cases, where the caller already knows that the netlink socket
 	// is in the same namespace as the calling thread, this can introduce a performance
 	// impact. In these cases, the caller can disable thread locking if performance
-	// consideration are of interest. If disabled, it is the responsibility of
+	// considerations are of interest. If disabled, it is the responsibility of
 	// the caller to make sure that all threads are running in the correct namespace.
-	// When DisabeNSLockThread is set, the caller cannot set the NetNS value. An error
+	// When DisableNSLockThread is set, the caller cannot set the NetNS value. An error
 	// will be returned.
 	DisableNSLockThread bool
 }
--- a/conn.go
+++ b/conn.go
@@ -583,4 +583,17 @@ type Config struct {
 	// CAP_SYS_ADMIN are required), and most applications should leave this set
 	// to 0.
 	NetNS int
+
+	// DisableNSLockThread
+	//
+	// By default, the library will lock the processing go routine to its corresponding
+	// thread in order to enable communication over netlink to a different namespaces.
+	// In some cases, where the caller already knows that the netlink socket
+	// is in the same namespace as the calling thread, this can introduce a performance
+	// impact. In these cases, the caller can disable thread locking if performance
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -583,4 +583,17 @@ type Config struct {
 	// CAP_SYS_ADMIN are required), and most applications should leave this set
 	// to 0.
 	NetNS int
+
+	// DisableNSLockThread
```
mdlayher: Nit: let's preserve the typical Go documentation style.

```
// DisableNSLockThread disables package netlink's default goroutine thread locking behavior.
// By default, the library will ...
```

And then you can continue on with what you have.

conn.go
```diff
@@ -583,4 +583,17 @@ type Config struct {
...
+	//
+	// By default, the library will lock the processing go routine to its corresponding
```
mdlayher:
```suggestion
	// By default, the library will lock the processing goroutine to its corresponding
```

conn.go
```diff
@@ -583,4 +583,17 @@ type Config struct {
...
+	// thread in order to enable communication over netlink to a different namespaces.
+	// In some cases, where the caller already knows that the netlink socket
+	// is in the same namespace as the calling thread, this can introduce a performance
+	// impact. In these cases, the caller can disable thread locking if performance
+	// consideration are of interest. If disabled, it is the responsibility of
```
mdlayher:
```suggestion
	// considerations are of interest. If disabled, it is the responsibility of
```

conn.go
```diff
@@ -583,4 +583,17 @@ type Config struct {
...
+	// the caller to make sure that all threads are running in the correct namespace.
+	// When DisabeNSLockThread is set, the caller cannot set the NetNS value. An error
```
mdlayher:
```suggestion
	// When DisableNSLockThread is set, the caller cannot set the NetNS value. An error
```

mdlayher: IMO just remove the last "An error will be returned" sentence.

conn_linux.go
```diff
@@ -15,6 +16,8 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+var errInvalidConfiguration = errors.New("netns cannot be set when disableNSThreadLock is set")
```
mdlayher: I think I have typically been prefixing these with "netlink:".

conn_linux.go
```diff
@@ -339,7 +342,7 @@ type sysSocket struct {
 // to a single thread.
 func newSysSocket(config *Config) (*sysSocket, error) {
 	// Determine network namespaces using the threadNetNS function.
-	g, err := newLockedNetNSGoroutine(config.NetNS, threadNetNS)
+	g, err := newLockedNetNSGoroutine(config.NetNS, threadNetNS, config.DisableNSLockThread)
```
mdlayher: Double negatives get a bit confusing later, so how about `!config.DisableNSLockThread` here so the called functions can use the positive form instead?

conn_linux.go
```diff
@@ -624,10 +627,16 @@ type lockedNetNSGoroutine struct {
 // newLockedNetNSGoroutine creates a lockedNetNSGoroutine that will enter the
 // specified network namespace netNS (by file descriptor), and will use the
 // getNS function to produce netNS handles.
-func newLockedNetNSGoroutine(netNS int, getNS func() (*netNS, error)) (*lockedNetNSGoroutine, error) {
+func newLockedNetNSGoroutine(netNS int, getNS func() (*netNS, error), disableThreadLock bool) (*lockedNetNSGoroutine, error) {
```
mdlayher: And then this would be `threadLock bool` or similar now.

conn_linux.go
mdlayher: Maybe name this `errNetNSDisabled` for clarity.

conn.go
dstiliadis: ok

conn.go
dstiliadis: yes

conn.go
dstiliadis: good point

conn_linux.go
dstiliadis: yes, I can see that. Pushing changes.

conn_linux.go
dstiliadis: good to keep consistency.

conn.go
dstiliadis: oops

dstiliadis:
Yes, we have done extensive testing. We use this package together with github.com/ti-mo/conntrack .. We see an average of 20% performance improvement in a 2-core processor when doing conntrack requests (reaches up to 8K conntrack operations per second).

dstiliadis:
Dealing with the double negatives gave me more trouble than expected :)  and I realized I needed go1.13 to enable this test. In any case, that should be cleaned up now.

Issue #130, opened 2019-03-18
netlink: OpError should implement Go 1.13+ errors.Wrapper interface


mdlayher:
This'll be straightforward, but here's a reminder to actually do it.

https://tip.golang.org/pkg/errors/#Wrapper

yarikk:
@mdlayher Oh, looks like you beat me on this…

PR #148, opened 2019-10-04
add IsExist() function

yarikk:
_IsExist_ comes handy when, say, checking the result of `RTM_NEWADDR`, and decide to proceed if the system failed to add address just because it's already there. Allows to skip scanning for existing addresses before attempting to add one, as the system shall do the same on its side.

cc @mdlayher @jsimonetti

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -44,6 +44,17 @@ func IsNotExist(err error) bool {
 	}
 }
 
+// IsExist is similar to IsNotExist but checks whether the error is known to report that an object already exists,
+// e.g. an address is already assigned to an interface.
+func IsExist(err error) bool {
+	switch err := err.(type) {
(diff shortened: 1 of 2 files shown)


mdlayher:
Thank you for the PR. As mentioned on Slack, I think it's worth investigating what should be done to integrate the new Go 1.13 errors and xerrors package APIs; see also https://github.com/mdlayher/netlink/issues/130.

If we can do that, we can also eliminate the IsNotExist helper.

yarikk:
Another PR shall follow.

mdlayher:
@yarikk check out #149. I'll merge shortly.

yarikk:
@mdlayher Oh, looks like you beat me on this…

Issue #123, opened 2019-02-21
netlink: release v1.0.0


mdlayher:
Once 1.12 hits and we're able to land runtime network poller support with proper timeouts, I think it's finally time to embrace semver and Go modules, and tag v1.0.0.

Before we do that, I'm debating if it's worth revising some previous interface decisions, such as:

- changing signature of `Config.Groups` and `Conn.Join/LeaveGroup` to pass `group int` rather than `group uint32`

This is passed as a 32-bit C `int` to the kernel, so Go's 64-bit `int` on 64-bit systems would need a sanity check.  `int` is generally a good default and the most common integer type in Go by far; `uint32` requires an explicit cast unless you're using an untyped constant (which is generally the case).

- unexporting or removing `Message.Marshal/UnmarshalBinary`
 
If you're importing this package, you're almost certainly using `Conn` which takes care of these internally. 
 We could also do faster and more clever things if we didn't have to worry about allocating for individual messages, or making a copy of byte slices during unmarshaling.

**DONE:** consider removing `HeaderType` and `HeaderFlags` prefixes from typed constants

I'm admittedly a little iffy on this, but in recent projects I tend to prefer more succinct names like `netlink.Request`, even if the type remains `Request HeaderFlags = 1`.  Is there a possibility of conflict in header types/flags with other current or future exported identifiers?  Would this be too confusing in calling code?

- finding another way to hide/unexport the `Socket` type and `NewConn` constructors

These are really only meant for nltest; I had considered doing unsafe `//go:linkname` at one point but can't remember why I didn't go with it.  It's not great, but it might be better than adding generally useless stuff to the exported 1.0.0 API.

**DONE:** - consider dropping `Conn.ReadWriteCloser`

This is only used in github.com/mdlayher/kobject as far as I know, and I think that its use there could be superseded by `Conn.SyscallConn` and methods there.

---

I think that about covers my thoughts. I know it's a lot, and I don't expect others to have strong opinions on all or any of these, but I would appreciate your feedback.  In addition, if there are other things you think are worth revisiting for 1.0.0, please do let me know.

I'm going to tag some of the regular contributors and users of this package in hopes that folks can help me work things out, although I encourage feedback from all who come across this issue.

/cc @florianl @ti-mo @terinjokes @jsimonetti @stapelberg @acln0

terinjokes:
> * finding another way to hide/unexport the `Socket` type and `NewConn` constructors

[...]

PR #149, opened 2019-10-04
netlink: make OpError compatible with errors.Is/Unwrap

### LINKED ISSUE #130: netlink: OpError should implement Go 1.13+ errors.Wrapper interface

mdlayher:

Fixes #130. Unfortunately, it probably isn't a good idea to rip out `netlink.IsNotExist` until Go 1.14/15 or so, when pretty much everyone can use errors.Is. Deprecate it.


mdlayher:
Updates #123 .

PR #151, opened 2019-10-05
netlink: add Nested method on Attribute{Decoder,Encoder}

mdlayher:

See the associated issue.

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -322,9 +322,9 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
 	}
 }
 
-// Nested decodes data into a nested AttributeDecoder and invokes fn repeatedly
-// for each attribute found in the nested data. When calling Nested, the Next
-// and Err methods should not be called on the nested AttributeDecoder.
+// Nested decodes data into a nested AttributeDecoder to handle nested netlink
+// attributes. When calling Nested, the Err method does not need to be called on
+// the nested AttributeDecoder.
 //
 // The nested AttributeDecoder nad inherits the same ByteOrder setting as the
 // top-level AttributeDecoder ad.
@@ -337,11 +337,8 @@ func (ad *AttributeDecoder) Nested(fn func(nad *AttributeDecoder) error) {
 		}
 		nad.ByteOrder = ad.ByteOrder
 
-		// Iterate until no more attributes or nad.Err() != nil.
-		for nad.Next() {
-			if err := fn(nad); err != nil {
-				return err
-			}
+		if err := fn(nad); err != nil {
+			return err
 		}
 
 		return nad.Err()
--- a/attribute.go
+++ b/attribute.go
@@ -189,6 +189,10 @@ func (ad *AttributeDecoder) Next() bool {
 
 // Type returns the Attribute.Type field of the current netlink attribute
 // pointed to by the decoder.
+//
+// Type masks off the high bits of the netlink attribute type which may contain
+// the Nested and NetByteOrder flags. If you need direct access to these flags,
+// consider using UnmarshalAttributes instead.
 func (ad *AttributeDecoder) Type() uint16 {
 	// Mask off any flags stored in the high bits.
 	return ad.attr().Type & attrTypeMask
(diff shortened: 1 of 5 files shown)


attribute.go
```diff
@@ -301,6 +307,32 @@ func (ad *AttributeDecoder) Do(fn func(b []byte) error) {
...
+	// Because we are wrapping Do, there is no need to check ad.err immediately.
+	ad.Do(func(b []byte) error {
+		nad, err := NewAttributeDecoder(b)
+		if err != nil {
+			return err
+		}
+		nad.ByteOrder = ad.ByteOrder
+
+		// Iterate until no more attributes or nad.Err() != nil.
+		for nad.Next() {
```
mdlayher: I'm not sure embedding the Next loop is a good idea because it makes certain operations harder. For example:

```go
// Array of peers.
			ad.Nested(func(nad *netlink.AttributeDecoder) error {
				// Individual peer fields are parsed from another nested set
				// of attributes within the peers array.
				var p wgtypes.Peer
				nad.Nested(func(nnad *netlink.AttributeDecoder) error {
					return parsePeer(nnad, &p)
				})

				// Only append this peer if no problems occurred.
				if nad.Err() != nil {
					return err
				}

				d.Peers = append(d.Peers, p)
				return nil
			})
```

The parsePeer function is now called repeatedly and only parses a single field at a time, which is a bit awkward.

attribute.go
```diff
@@ -185,7 +190,8 @@ func (ad *AttributeDecoder) Next() bool {
 // Type returns the Attribute.Type field of the current netlink attribute
 // pointed to by the decoder.
 func (ad *AttributeDecoder) Type() uint16 {
-	return ad.attr().Type
+	// Mask off any flags stored in the high bits.
+	return ad.attr().Type & attrTypeMask
```
florianl: Hiding a part of the original Type might be dangerous. I think, user should decide by themselves, if they want to see only parts of the Type.

attribute.go
florianl: it might also lead to hard to read code, if there are multiple level of nestings.

e.g. in case of conntrack, the attribute CTA_IP_V6_SRC is nested in CTA_TUPLE_IP, which is nested in CTA_TUPLE_ORIG.

message.go
```diff
@@ -7,6 +7,16 @@ import (
 	"github.com/mdlayher/netlink/nlenc"
 )
 
+// Flags which may apply to netlink attribute types when communicating with
+// certain netlink families.
+const (
+	Nested       = 0x8000
+	NetByteOrder = 0x4000
```
florianl: I really like these two exports :)

attribute.go
mdlayher: I don't think this is dangerous because these bits are used by the kernel to handle the Nested and NetByteOrder flags. If you want direct access to the full bits, you can use UnmarshalAttributes as normal.

attribute.go
mdlayher: I changed this so that the caller is still responsible for calling ad.Next. As far as multiple levels of nesting, I think it's up to the caller to use good judgment and refactor out heavily indented blocks.

attribute.go
florianl: what about making this masking optional? Or providing second function without masking?

So far, it was handy to use `ad.Type()` in a switch statement and then cover the resulting types.

mdlayher: I think my concern with exposing the flag bits at all is that the caller will have to remember to account for them, or make a duplicate case such as https://github.com/WireGuard/wgctrl-go/commit/dc235175870abf30dabaac70a81a99f479e99178#diff-2bfaf714bbb2315a74c7760755088e00L78.

What's the advantage to leaving them exposed? If you are implementing a genetlink interface, you probably have docs in front of you that indicate that a certain attribute contains nested attributes. We don't rely on the presence of the flag to switch to a different nested mode; it's all up to the caller to invoke ad.Nested and such as appropriate.

I just want to understand your use case. I would rather not add another function or an argument to mask or not.

mdlayher: It would appear that libnl also masks off these bits: https://www.infradead.org/~tgr/libnl/doc/api/attr_8c_source.html#l00109.

PR #153, opened 2019-10-07
Add NLA_FLAG

florianl:

This PR adds the attribute type NLA_FLAG.

NLA_FLAG does not hold data. It is used in the net/tipc netlink family.

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -282,11 +282,7 @@ func (ad *AttributeDecoder) Uint64() uint64 {
 
 // Flag returns a boolean representing the Attribute.
 func (ad *AttributeDecoder) Flag() bool {
-	if ad.err != nil {
-		return false
-	}
-
-	return true
+	return ad.err == nil
 }
 
 // Do is a general purpose function which allows access to the current data
@@ -398,7 +394,7 @@ func (ae *AttributeEncoder) Flag(typ uint16, v bool) {
 	if ae.err != nil {
 		return
 	}
-	if v == false {
+	if !v {
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


attribute.go
```diff
@@ -280,6 +280,11 @@ func (ad *AttributeDecoder) Uint64() uint64 {
 	return ad.ByteOrder.Uint64(b)
 }
 
+// Flag returns a boolean representing the Attribute.
+func (ad *AttributeDecoder) Flag() bool {
+	return ad.err == nil
```
mdlayher:
```suggestion
	// If Flag is called and there is no current error, assume that the caller encountered
	// a flag attribute. The presence of a flag attribute means this should always return true.
	return ad.err == nil
```

attribute.go
```diff
@@ -384,6 +389,18 @@ func (ae *AttributeEncoder) Uint64(typ uint16, v uint64) {
...
+// Flag encodes a flag into an Attribute specidied by typ.
+func (ae *AttributeEncoder) Flag(typ uint16, v bool) {
+	if ae.err != nil {
+		return
+	}
+	if !v {
+		return
+	}
+
+	ae.attrs = append(ae.attrs, Attribute{Type: typ})
```
mdlayher:
```suggestion
	// Length and Data are unused for flag attributes.
	ae.attrs = append(ae.attrs, Attribute{Type: typ})
```

attribute.go
```diff
@@ -384,6 +389,18 @@ func (ae *AttributeEncoder) Uint64(typ uint16, v uint64) {
 	})
 }
 
+// Flag encodes a flag into an Attribute specidied by typ.
+func (ae *AttributeEncoder) Flag(typ uint16, v bool) {
+	if ae.err != nil {
```
mdlayher: Nit: `if ae.err != nil || !v` as a slight simplification.

mdlayher:
Merged and will fix a couple nits myself, thanks!

PR #156, opened 2019-12-05
netlink: v1.0.0

### LINKED ISSUE #123: netlink: release v1.0.0

mdlayher:

Fixes #123.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,16 +12,15 @@ channel!
 
 ## Stability
 
-At this time, package `netlink` is in a pre-v1.0.0 state. Changes are being made
-which may impact the exported API of this package and others in its ecosystem.
-To follow along on the status of a v1.0.0 release, [see the associated issue](https://github.com/mdlayher/netlink/issues/123).
+This package has reached v1.0.0 and any future breaking API changes will prompt
+the release of a new major version. Features and bug fixes will continue to
+occur in the v1.x.x series.
 
 The general policy of this package is to only support the latest, stable version
 of Go. Compatibility shims may be added for prior versions of Go on an as-needed
 basis. If you would like to raise a concern, please [file an issue](https://github.com/mdlayher/netlink/issues/new).
 
-**If you depend on this package in your applications, please vendor it or use Go
-modules when building your application.**
+**If you depend on this package in your applications, please use Go modules.**
 
 ## Design
 


mdlayher:
Package netlink is already widely used in many production applications, so this v1.0.0 is more of a symbolic first release, as well as a committment to stability for the v1.x.x series.

Thank you to all of the netlink contributors and users for your support over the past few years!

Issue #115, opened 2018-11-14
netlink: consider walking attributes slice directly in AttributeDecoder


mdlayher:
As a first step right now, the `NewAttributeDecoder` constructor unpacks the attributes from the byte slice so it can walk them internally.  It'd be interesting to try to directly walk the byte slice instead, to avoid additional internal allocations and copies.

This could also mean that the constructor could stop returning an error and an invalid attributes slice could be caught by the `Err` method instead.

mdlayher:
This is probably not necessary, so closing.

PR #157, opened 2019-12-06
Opportunistically decode attributes in AttributeDecoder

ti-mo:
Hi Matt,

As discussed yesterday, this patch contributes two things:

- adds `AttributeDecoder.TypeFlags()` that returns the Type with an inverted `attrTypeMask`
- inverts the call chain between `UnmarshalAttributes()` and `ad.Next()`

This change is motivated by https://github.com/ti-mo/conntrack/issues/13, as I'd like to expose a `netfilter.AttributeDecoder/Encoder` and `conntrack.AttributeDecoder/Encoder` interface all the way up the call chain. If done right, decoding a whole netlink message can be done with only a single allocation of each layer's `Attribute` type. :scream: 

It doesn't save much on the benches in the repo, because it decodes all attributes anyway, but it allows for large efficiency gains in downstream libraries. Because attributes can now be decoded opportunistically, (eg. calling `ad.Next()` until you have the attributes you care about and calling it quits), a caller that uses AttributeDecoder will no longer implicitly decode the whole message. This potentially saves a lot of CPU time.

I also cut down on memory allocations, since AttributeDecoder now uses a single `netlink.Attribute` as a scratch buffer it repeatedly unmarshals into, completely removing allocations from the hot path.

This is illustrated by the benchmark results.

Before:
```
λ  netlink 38c649d ✓ go test -bench BenchmarkUnmarshalAttributes
goos: linux
goarch: amd64
pkg: github.com/mdlayher/netlink
BenchmarkUnmarshalAttributes/0-16  	354637596	         3.13 ns/op	       0 B/op	       0 allocs/op
BenchmarkUnmarshalAttributes/1-16  	 8755629	       134 ns/op	      33 B/op	       2 allocs/op
BenchmarkUnmarshalAttributes/8-16  	 1403520	       856 ns/op	     544 B/op	      12 allocs/op
BenchmarkUnmarshalAttributes/64-16 	  182522	      6417 ns/op	    8160 B/op	      71 allocs/op
BenchmarkUnmarshalAttributes/512-16         	   10000	    114465 ns/op	  294880 B/op	     522 allocs/op
```

After:
```
λ  netlink master ✓ go test -bench BenchmarkUnmarshalAttributes
goos: linux
goarch: amd64
pkg: github.com/mdlayher/netlink
BenchmarkUnmarshalAttributes/0-16  	13132034	        92.4 ns/op	     112 B/op	       1 allocs/op
BenchmarkUnmarshalAttributes/1-16  	 5177278	       234 ns/op	     145 B/op	       3 allocs/op
BenchmarkUnmarshalAttributes/8-16  	 1474484	       806 ns/op	     432 B/op	      10 allocs/op
BenchmarkUnmarshalAttributes/64-16 	  183102	      6580 ns/op	    6256 B/op	      66 allocs/op
BenchmarkUnmarshalAttributes/512-16         	   10000	    103788 ns/op	  278640 B/op	     514 allocs/op
```

The extra allocation visible in 0 and 1 is due to an added `make([]Attribute, 0, ad.Len())` in `UnmarshalAttributes` that pre-allocates the backing array of the output slice. This amortizes nicely over time.

The existing test suite was very helpful in weeding out any inconsistencies in behaviour, so for example I had to add an extra check in `UnmarshalAttributes` to make sure it explicitly returns a nil `[]Attribute` when no attributes where decoded.

Let me know what you think!

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -107,26 +107,27 @@ func MarshalAttributes(attrs []Attribute) ([]byte, error) {
 // It is recommend to use the AttributeDecoder type where possible instead of calling
 // UnmarshalAttributes and using package nlenc functions directly.
 func UnmarshalAttributes(b []byte) ([]Attribute, error) {
-	var attrs []Attribute
-	var i int
-	for {
-		if i > len(b) || len(b[i:]) == 0 {
-			break
+
+	ad, err := NewAttributeDecoder(b)
+	if err != nil {
+		return nil, err
+	}
+
+	// Return a nil slice when there are no attributes to decode.
+	if ad.Len() == 0 {
+		return nil, nil
+	}
+
+	attrs := make([]Attribute, 0, ad.Len())
+
+	for ad.Next() {
+		if ad.attr().Length != 0 {
+			attrs = append(attrs, ad.attr())
 		}
+	}
 
-		var a Attribute
-		if err := (&a).unmarshal(b[i:]); err != nil {
-			return nil, err
-		}
-
-		if a.Length == 0 {
-			i += nlaHeaderLen
-			continue
-		}
-
-		i += nlaAlign(int(a.Length))
-
-		attrs = append(attrs, a)
+	if err := ad.Err(); err != nil {
+		return nil, err
 	}
 
 	return attrs, nil
@@ -148,10 +149,14 @@ type AttributeDecoder struct {
 	// If not set, the native byte order will be used.
 	ByteOrder binary.ByteOrder
 
-	// The attributes being worked on, and the iterator index into the slice of
-	// attributes.
-	attrs []Attribute
-	i     int
+	// The current attribute being worked on.
+	a Attribute
+
+	// The slice of input bytes and its iterator index.
+	b []byte
+	i int
+
+	len int
 
 	// Any error encountered while decoding attributes.
 	err error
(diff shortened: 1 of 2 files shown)


attribute.go
```diff
@@ -107,26 +107,27 @@ func MarshalAttributes(attrs []Attribute) ([]byte, error) {
...
+	// Return a nil slice when there are no attributes to decode.
+	if ad.Len() == 0 {
+		return nil, nil
+	}
 
-		i += nlaAlign(int(a.Length))
+	attrs := make([]Attribute, 0, ad.Len())
+
+	for ad.Next() {
+		if ad.attr().Length != 0 {
```
mdlayher: I seem to remember some weirdness around flag attributes where they might have length 0. I don't think we should filter these probably. /cc @florianl

ti-mo: Let's follow up in an issue as discussed: https://github.com/mdlayher/netlink/blob/master/attribute_test.go#L285

As it stands, `UnmarshalAttributes` does not emit any zero-length attributes, so this clause was added to match existing behaviour.

attribute.go
```diff
@@ -181,34 +190,93 @@ func (ad *AttributeDecoder) Next() bool {
...
+	return true
 }
 
 // Type returns the Attribute.Type field of the current netlink attribute
 // pointed to by the decoder.
 //
 // Type masks off the high bits of the netlink attribute type which may contain
-// the Nested and NetByteOrder flags. If you need direct access to these flags,
-// consider using UnmarshalAttributes instead.
+// the Nested and NetByteOrder flags. These can be obtained by calling TypeFlags().
```
mdlayher: Nit: drop the parens on the doc. I think we go without them elsewhere.

attribute.go
```diff
@@ -181,34 +190,93 @@ func (ad *AttributeDecoder) Next() bool {
...
+	return ad.a.Type & ^attrTypeMask
 }
 
 // Len returns the number of netlink attributes pointed to by the decoder.
-func (ad *AttributeDecoder) Len() int { return len(ad.attrs) }
+func (ad *AttributeDecoder) Len() int { return ad.length }
+
+// count scans the input slice to count the number of netlink attributes
+// that could be decoded by Next().
+func (ad *AttributeDecoder) count() (int, error) {
```
mdlayher: I think that having both count and Len is a bit confusing.

How about "available"?

ti-mo:
> LGTM overall, just a couple of nits. I assume your fuzzing is going well?

At about an hour now, had to interrupt but resumed from the same corpus.

`2019/12/06 16:00:16 workers: 16, corpus: 76 (2m46s ago), crashers: 0, restarts: 1/9994, execs: 301119156 (149143/sec), cover: 87, uptime: 33m39s`

Thanks for the review!

ti-mo:
Fuzzer ran for 4.5hrs without any crashers. :point_up:

ti-mo:
Looks like this implements https://github.com/mdlayher/netlink/issues/115!

attribute.go
florianl: i've run tests for q_dsmark (where attribute type flag is used) with this PR and all looks good :+1:

Issue #136, opened 2019-05-04
netlink: blocking Receive prevents Send


acln0:
I encountered this while using the nfqueue subsystem. Given a `var nlconn *netlink.Conn`:

One goroutine (henceforth "the producer") does approximately this:

```
for {
	msgs, err := nlconn.Receive()
	if err != nil {
		// handle	
	}
	ch <- msgs
}
```

It sends the messages to a channel to fan them out to workers. The workers do approximately this:

```
for msgs := range ch {
	for _, msg := range msgs {
		reply := process(msg)
		nlconn.Send(reply)
	}
}
```

To set the stage for the pseudo-deadlock, assume the producer calls `Receive`. The request travels down to the `lockedNetNSGoroutine` associated with the connection, it is sent on `g.funcC`.

So far so good. The `lockedNetNSGoroutine` receives the request, runs it, and blocks [here](https://github.com/mdlayher/netlink/blob/b540351f6c5108aa445eaa6677010703c7294f8d/conn_linux.go#L647).

Meanwhile, a worker has finished processing a message it was handed previously, and wants to send a reply. It calls `Send`, which blocks [here](https://github.com/mdlayher/netlink/blob/b540351f6c5108aa445eaa6677010703c7294f8d/conn_linux.go#L672) because the `lockedNetNSGoroutine` is busy waiting for the blocking call to `recvmsg` to return, and is not servicing the channel.

I have refactored my code such that all the processing happens synchronously for now, so instead of a producer and a number of workers, I have only this:

```
for {
	msgs, err := nlconn.Receive()
	if err != nil {
		// handle
	}

	for msgs := range ch {
		for _, msg := range msgs {
			reply := process(msg)
			nlconn.Send(reply)
		}
	}
}
```

This works, and I don't think it is prohibitively expensive in terms of performance. It may even be faster than fanning out. I haven't measured anything yet, since my project is in a prototype stage. There may also be the possibility for me to use two different `*netlink.Conn`s bound to the same netfilter queue, one for receiving and one for sending. I don't know if this works: I have not tried it yet.

I thought about potential solutions for a little while, but I couldn't figure out any good ones. Therefore, I have filed this bug for discussion. Can something be done here? Should we at least document this? While debugging the issue, I was stumped for a while, since `Receive` and `Send` both acquire a read lock (!), and it seemed absurd to me that they'd be mutually exclusive in a deeper sense, until I saw the stack traces of the goroutines involved.

Thanks,
Andrei

[...]

PR #160, opened 2020-01-05
Add select

alexispires:
Allow ``select`` to this kind of design:

```go
for i := 1; i > 0; i, _ = conn.Select(&unix.Timeval{}) {
	rmsg, err := conn.Receive()

	if err != nil {
		return fmt.Errorf("Receive: %w", err)
	}

	for _, msg := range rmsg {
		// do something
	}
}
```

I need this to trying reproduce behaviors of nftables which check with ``select`` before doing a ``recvmsg``, example of strace:

```bash
select(4, [3], NULL, NULL, {tv_sec=0, tv_usec=0}) = 1 (in [3], left {tv_sec=0, tv_usec=0})
recvmsg(......)
select(4, [3], NULL, NULL, {tv_sec=0, tv_usec=0}) = 1 (in [3], left {tv_sec=0, tv_usec=0})
recvmsg(......)
select(4, [3], NULL, NULL, {tv_sec=0, tv_usec=0}) = 0 (Timeout)
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -49,6 +49,7 @@ type Socket interface {
 	Send(m Message) error
 	SendMessages(m []Message) error
 	Receive() ([]Message, error)
+	Select() (int, error)
 }
 
 // Dial dials a connection to netlink, using the specified netlink family.
@@ -317,6 +318,11 @@ func (c *Conn) receive() ([]Message, error) {
 			return res, nil
 		}
 	}
+}
+
+// Select allow to check whether netlink messages are available
+func (c *Conn) Select() (int, error) {
+	return c.sock.Select()
 }
 
 // A groupJoinLeaver is a Socket that supports joining and leaving
--- a/conn.go
+++ b/conn.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"golang.org/x/net/bpf"
+	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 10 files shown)


mdlayher:
I think it makes more sense to use an API like SyscallConn for this, although I'm also not sure why you wouldn't just let receive wait on the socket since the runtime handles it in an efficient way.

https://godoc.org/github.com/mdlayher/netlink#Conn.SyscallConn

alexispires:
> I think it makes more sense to use an API like SyscallConn for this, although I'm also not sure why you wouldn't just let receive wait on the socket since the runtime handles it in an efficient way.
> 
> https://godoc.org/github.com/mdlayher/netlink#Conn.SyscallConn

Ok it works by using SyscallConn.

I need this to prevent to be blocked by recvmsg because the func which make the syscall need to return some results. As nftables:
http://git.netfilter.org/nftables/tree/src/mnl.c#n349

I think it's better to have it directly in this lib but you can close this PR if you want ;)

alexispires:
Related to https://github.com/google/nftables/pull/88

mdlayher:
It could be worth pursuing something like this as an internal change to unblock issues like #136, but I don't want to expose this API directly. Will close this, but thanks for your time.

PR #161, opened 2020-02-05
netlink: handle EINTR when checking for syscall readiness

mdlayher:

Per the tentative 1.14 release notes, we must be prepared to deal with EINTR on slow system calls, such as a blocking read waiting on a multicast group.

https://tip.golang.org/doc/go1.14#runtime

I'm pretty sure this is all we have to do, but I'd appreciate more eyes on this.


acln0:
I think it's very hard to write a test case to exercise this new code path, unfortunately. I also think that the code is correct, so there isn't such a pressing need for a test.

I remember https://github.com/golang/go/issues/20400, but this is about something slightly different, though somewhat related, I guess.

florianl:
I agree with @acln0 and so it looks good.

mdlayher:
Yeah I also have no idea how to test that behavior, but also didn't think it was super necessary for this case. Thanks!

Issue #164, opened 2020-03-24
Improve behaviour of SetReadBuffer / SetWriteBuffer


MarkusBauer:
Hi, 
after dealing with netlink socket's buffer size I compared mdlayher/netlink's implementation of `SetReadBuffer` with libnfnetlink's `nfnl_rcvbufsiz` and found an improvement:

[SetReadBuffer](https://github.com/mdlayher/netlink/blob/master/conn_linux.go#L278) uses a single syscall with `SO_RCVBUF` to set the buffer size. 
In contrast [libnfnetlink's nfnl_rcvbufsiz](https://github.com/threatstack/libnfnetlink/blob/master/src/libnfnetlink.c#L1012) uses `SO_RCVBUFFORCE` with a fallback to `SO_RCVBUF` if that fails.

This difference is important if the user requests a buffer size that is larger than the system's limit on buffer sizes. `SO_RCVBUFFORCE` ignores this limit if the user is root (or has `CAP_NET_ADMIN` set), while `SO_RCVBUF` fails (see [man socket](https://linux.die.net/man/7/socket) for details). In my use case, this difference would save me from tampering with global limits if I need a large netlink buffer.

Implementing SetReadBuffer with `SO_RCVBUFFORCE` will allow users to fully use their privileges without limiting the abilities of unprivileged users. Given that buffer sizes are typically changed only once performance impact should be negligible. 

If you approve this change I'll try to implement this and send you a PR.

mdlayher:
Sounds good to me, please do open a PR when you can. Thanks!

mdlayher:
Closed by #165 .

PR #165, opened 2020-03-24
Improve behaviour of SetReadBuffer / SetWriteBuffer

MarkusBauer:
(see issue #164 for a detailed explanation)

This PR allows users to use their privileges regarding buffer sizes: root and users with `CAP_NET_ADMIN` can ignore the system's limit on buffer sizes. Unprivileged users are not affected, there's a fallback. The testcase has been adjusted.

While I was working on this PR I noticed that the library has no method to read back the actual buffer size of the socket. I added these methods (`GetReadBuffer` and `GetWriteBuffer`) which are similar to their Write pendants.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -276,21 +276,41 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
 // SetReadBuffer sets the size of the operating system's receive buffer
 // associated with the Conn.
 func (c *conn) SetReadBuffer(bytes int) error {
-	return os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
+	// First try SO_RCVBUFFORCE. Given necessary permissions this syscall ignores limits.
+	err := os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
 		unix.SOL_SOCKET,
-		unix.SO_RCVBUF,
+		unix.SO_RCVBUFFORCE,
 		bytes,
 	))
+	if err != nil {
+		// If SO_SNDBUFFORCE fails, try SO_RCVBUF
+		err = os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
+			unix.SOL_SOCKET,
+			unix.SO_RCVBUF,
+			bytes,
+		))
+	}
+	return nil
 }
 
 // SetReadBuffer sets the size of the operating system's transmit buffer
 // associated with the Conn.
 func (c *conn) SetWriteBuffer(bytes int) error {
-	return os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
+	// First try SO_SNDBUFFORCE. Given necessary permissions this syscall ignores limits.
+	err := os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
 		unix.SOL_SOCKET,
-		unix.SO_SNDBUF,
+		unix.SO_SNDBUFFORCE,
 		bytes,
 	))
+	if err != nil {
+		// If SO_SNDBUFFORCE fails, try SO_SNDBUF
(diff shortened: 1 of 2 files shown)


mdlayher:
Hey, my apologies for the delay here, I totally lost track of this. I will review now.

conn_linux.go
```diff
@@ -276,21 +277,67 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
...
-	return os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
+	// First try SO_RCVBUFFORCE. Given necessary permissions this syscall ignores limits.
+	err := os.NewSyscallError("setsockopt", c.s.SetSockoptInt(
 		unix.SOL_SOCKET,
-		unix.SO_RCVBUF,
+		unix.SO_RCVBUFFORCE,
 		bytes,
 	))
+	if err != nil {
+		// If SO_SNDBUFFORCE fails, try SO_RCVBUF
```
mdlayher, line 287: Is there a specific syscall error we could inspect to determine if fallback is needed? Would something like `os.IsPermission` be sufficient?

conn_linux.go
```diff
@@ -276,21 +277,67 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
...
+
+// GetReadBuffer retrieves the size of the operating system's receive buffer
+// associated with the Conn.
+func (c *conn) GetReadBuffer() (int, error) {
+	value, err := c.s.GetSockoptInt(
+		unix.SOL_SOCKET,
+		unix.SO_RCVBUF,
+	)
+	if err != nil {
+		return value, os.NewSyscallError("getsockopt", err)
```
mdlayher, line 325: Return 0 to be more explicit in error cases.

conn_linux.go
```diff
@@ -276,21 +277,67 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
...
+	}
+	return value, err
```
mdlayher, line 327: Return nil error.

conn_linux.go
```diff
@@ -276,21 +277,67 @@ func (c *conn) SetWriteDeadline(t time.Time) error {
...
+}
+
+// GetWriteBuffer retrieves the size of the operating system's transmit buffer
+// associated with the Conn.
+func (c *conn) GetWriteBuffer() (int, error) {
```
mdlayher, line 332: Same comments.

conn_linux.go
```diff
@@ -597,6 +644,19 @@ func (s *sysSocket) SetSockoptInt(level, opt, value int) error {
 	return err
 }
 
+func (s *sysSocket) GetSockoptInt(level, opt int) (int, error) {
+	var value int
```
mdlayher, line 648: Nit:
```go
var (
    value int
    err error
)
```

conn_linux.go
```diff
@@ -597,6 +644,19 @@ func (s *sysSocket) SetSockoptInt(level, opt, value int) error {
 	return err
 }
 
+func (s *sysSocket) GetSockoptInt(level, opt int) (int, error) {
```
mdlayher, line 647: Same comments about return values being explicitly 0 or nil.

conn_linux.go
MarkusBauer: The man page does not require a specific error for that case, but the kernel uses `EPERM` (which is the only logical one of course). Also the permission check is the first thing to happen. I don't think it will ever make a difference if we check for permission errors or not, `libnfnetlink` does not check for it. Your choice.

MarkusBauer:
Thank you for your review. I adressed your remarks.

conn_linux.go
mdlayher: I'm good with leaving it be then, thanks for the info.

PR #169, opened 2020-06-10
netlink: Conn.Close unblocks Conn.Receive

### LINKED ISSUE #162: netlink: Close does not unblock concurrent Receive operation
Similar to https://github.com/mdlayher/netlink/issues/136, though the use-case is different.
I would want to listen for any IP Route changes so I use blocking Receive() for multicast messages. I cannot use timeout on it because than there is a chance to miss events while restarting the Receive cycle.
In order to break the Receive (i.e. on graceful exit), I close the Conn on another goroutine, which appears to be the only way to do it.
The problem is that Receive in it's core read function acquires the R side of the RWMutex:
https://github.com/mdlayher/netlink/blob/ad88bf8eff5c1fd8d6a29299addeeb23d05f1ec1/conn_linux.go#L369
Close, however wants to acquire the W side here:
https://github.com/mdlayher/netlink/blob/ad88bf8eff5c1fd8d6a29299addeeb23d05f1ec1/conn_linux.go#L502
which is not possible until the R(eader) side is freed.
This results in a deadlock, until something is received on the socket, which is not deterministic.
I don't have a solution in mind, but maybe close function should lock the W side of the mutex only after calling the close syscall, ensuring all blocking operations are woke up (and so releasing the lock). If other parallel operations are handling the errors appropriately, it might not cause problems if `s.closed` temporarily appears to be false, while the socket is already closed.

mdlayher:
I believe this will fix #162, but I would appreciate a careful review to make sure I'm not missing anything.

/cc @acln0 and @libesz @bradfitz @MXWXZ who have run into this bug

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -6,7 +6,7 @@ import (
 	"math"
 	"os"
 	"runtime"
-	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
@@ -86,6 +86,7 @@ func dial(family int, config *Config) (*conn, uint32, error) {
 		defer threadNS.Restore()
 	}
 
+	// Socket will establish the internal state of the sysSocket structure.
 	sock := &sysSocket{}
 	if err := sock.Socket(family); err != nil {
 		return nil, 0, os.NewSyscallError("socket", err)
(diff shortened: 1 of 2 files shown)


mdlayher:
With Go 1.11 finally put to bed, I've rebased this and all tests seem to pass. I'm going to merge this with the intent of fixing #162 but would appreciate more eyes on this change if anything seems amiss.

Issue #154, opened 2019-10-09
netlink: performance bottleneck due to single goroutine in high performance applications


smarterclayton:
In scenarios where a locked OS thread isn't necessary, is the goroutine in sysSocket.write/read still required?  Perhaps I'm missing a subtlety of M:N with the use of Recvmsg and netlink, but my expectation was that with the proper descriptor any goroutine could read correctly from that socket if the netns is consistent.  Is that not the case?

I ask because in high traffic systems I'm seeing a fair amount of go runtime scheduler time in profiles when GOMAXPROCS>1 in go 1.12 due to the forced go routine change, and in testing in lightly loaded setups with a 100-200 netlink multicast event per second rate I saw a 5-10% CPU improvement if the goroutine is skipped (some from not needing to close the channel or defer).  I am not a socket expert of course in this case so I figured it was easier to ask.

mdlayher:
Are you setting DisableNSLockThread? If so I would be interested in seeing what optimizations can be made on that path and would welcome PRs. I haven't had a need for higher performance myself yet.

If you truly need maximum performance, you could also use the SyscallConn method and deal with the socket directly.

smarterclayton:
In this case yes - this is a privileged per host conntrack monitor which is
in the root net namespace.

mdlayher:
I'm back on my computer today and had a chance to think about this more.

First, is your code open source? That'd help me understand what all it is that you're doing.

Second, the original reasons for the goroutine locking are here:

https://github.com/mdlayher/netlink/blob/99e7bba29c9d8ed375750e1531f580f3190cfb60/conn_linux.go#L655-L674

I suspect that at the time, I may have had another bug that caused issues with routing messages between sockets. If the caller disables thread locking via the config, perhaps we can allow multiple goroutines to directly access the socket simultaneously without any issues. That said, I haven't bothered to challenge the status quo as my needs are a lot simpler than many of the nfnetlink use cases that have popped up with folks like yourself, @ti-mo, and @florianl.

If we can prove that it's safe to do so when network namespace manipulation is not required, I'd love to see a free performance boost for folks who don't need that functionality.

ti-mo:
I've never personally encountered the issue @mdlayher describes about messages being routed to the wrong sockets, always been wondering what the exact symptoms would be. I suspect it would look like the situation I'm describing below if network namespaces were involved.

[...]

PR #171, opened 2020-07-28
Reduce thread locking behaviour to dial phase only

ti-mo:
:wave: Back again with another controversial change. :smile: This aims to eliminate the bottleneck described in https://github.com/mdlayher/netlink/issues/154. This is the approach I described in https://github.com/mdlayher/netlink/issues/154#issuecomment-578045314. With the goal of dropping support for Go 1.11 and lower, this might be a good time to revisit.

To Do:
- [ ] Reintroduce a test for machines with network namespaces disabled, since `newLockedNetNSGoroutine` was removed

---

There are 2 elements at play here:
- The message misrouting discussed in http://lists.infradead.org/pipermail/libnl/2017-February/002293.html
  With this change, I'd like to start spelunking older versions of the kernel and the Go runtime for this bug. @mdlayher are you positive the `TestIntegrationConnConcurrent*` family of tests is able to reproduce this? Can you remember which Go and kernel version you were using at the time when this was written?

- Socket creation inside network namespaces
  As far as namespaces are concerned, OS threads only briefly need to enter a netns to create the netlink socket and can safely be returned to the host namespace once the socket fd has been allocated. We don't need to lock OS threads permanently to operate in multiple namespaces.

---

Some feedback please. :pray:  Still looking for a way to inject a mock `func getNS` somewhere into the `dial()` code path to emulate systems with network namespaces disabled.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -589,21 +589,6 @@ type Config struct {
 	// to 0.
 	NetNS int
 
-	// DisableNSLockThread disables package netlink's default goroutine thread
-	// locking behavior.
-	//
-	// By default, the library will lock the processing goroutine to its
-	// corresponding OS thread in order to enable communication over netlink to
-	// a different network namespace.
-	//
-	// If the caller already knows that the netlink socket is in the same
-	// namespace as the calling thread, this can introduce a performance
-	// impact. This option disables the OS thread locking behavior if
-	// performance considerations are of interest.
-	//
-	// If disabled, it is the responsibility of the caller to make sure that all
-	// threads are running in the correct namespace.
-	//
-	// When DisableNSLockThread is set, the caller cannot set the NetNS value.
+	// DisableNSLockThread is deprecated and has no effect.
 	DisableNSLockThread bool
 }
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -3,7 +3,6 @@
 package netlink
 
 import (
-	"errors"
 	"math"
 	"os"
 	"runtime"
@@ -46,7 +45,7 @@ type socket interface {
 	GetSockoptInt(level, opt int) (int, error)
 }
 
-// dial is the entry point for Dial.  dial opens a netlink socket using
+// dial is the entry point for Dial. dial opens a netlink socket using
 // system calls, and returns its PID.
 func dial(family int, config *Config) (*conn, uint32, error) {
 	// Prepare sysSocket's internal loop and create the socket.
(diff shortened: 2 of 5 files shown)


conn.go
```diff
@@ -589,21 +589,6 @@ type Config struct {
...
-	// If the caller already knows that the netlink socket is in the same
-	// namespace as the calling thread, this can introduce a performance
-	// impact. This option disables the OS thread locking behavior if
-	// performance considerations are of interest.
-	//
-	// If disabled, it is the responsibility of the caller to make sure that all
-	// threads are running in the correct namespace.
-	//
-	// When DisableNSLockThread is set, the caller cannot set the NetNS value.
+	// DisableNSLockThread is deprecated.
```
mdlayher, line 592: How about:

```
// DisableNSLockThread is deprecated and has no effect.
```

ti-mo:
Sorry, I haven't been around to work on this. Initial results are promising! The integration test suite completes about twice as fast on my machine, reflected in the profiles by an absence of lock contention around the worker closure channel. (since it's gone, obviously)

I'll do some virtme spelunking against older kernels and see if I can reproduce this thing. Take care!

ti-mo:
Alright, I went as far back as kernel `3.2.0-5-amd64` that shipped with Debian Wheezy, integration test suite ran without any hiccups. Test binaries built with `go1.15.2`. All is well!

ti-mo:
Timing the execution of the test suite on an 8-core Ryzen 3700X running kernel `5.7.19-1-ck`. Test binary built with go1.15.2.

Before: `sudo ./netlink.test -test.v  29.23s user 21.35s system 1299% cpu 3.892 total`
After: `sudo ./netlink.test -test.v  4.29s user 0.79s system 532% cpu 0.955 total`

These are not one-off results, the execution time sat consistently around the values shown.

Compatibility-wise, the integration test suite passes on systems as early as Linux `3.2.0-5-amd64`
with test binaries built with Go 1.10.

ti-mo:
> To clarify, would you intend for this to go into a v1.2.0 (Go 1.12+ only) and _not_ v1.1.1 (Go 1.11+)?

According to the test suite, this works with Go 1.10 just fine, so not sure if we should worry too much about this.

Re: mocking a test for disabled namespaces: `fileNetNS()` doesn't have tests right now, which we should probably address. Still thinking about how to test `Dial()`. Any ideas?

ti-mo:
@mdlayher I've added a short `netns_linux_test.go` to validate the behaviour of `netns.disabled`. Along with the existing integration test suite, that brings coverage of `netns_linux.go` to 100%. Feel free to merge if no more feedback from your side!

mdlayher:
Thanks again for this.

PR #173, opened 2020-10-12
netlink: drop support for Go 1.11

### LINKED ISSUE #170: netlink: drop support for Go 1.11 and below at Go 1.15 release
Go 1.12 was released in February 2019, and enabled us to integrate with the runtime network poller for deadline and SyscallConn support.

With Go 1.15 on the horizon, I'd like to officially drop support for Go 1.11 and below when I tag v1.2.0. This means that I will incorporate the few minor changes since v1.1.0 in a v1.1.1, but that version will be the last to support Go 1.11 and below.

I understand some users are stuck on LTS versions of Go shipped with their operating systems, but they will still be able to use v1.1.1 indefinitely until they can upgrade. The Go project policy is to only support the latest two versions of Go, which as of 1.15 would be 1.14 and 1.15. I think it's time to make a clean break and remove a lot of build tag shims.

mdlayher:

Fixes #170. Rejoice!

/cc @ti-mo


conn_linux.go
```diff
@@ -463,11 +470,6 @@ func (s *sysSocket) Socket(family int) error {
 	// For a more thorough explanation, see similar work in the
 	// Go tree: func sysSocket in net/sock_cloexec.go, as well
 	// as the detailed comment in syscall/exec_unix.go.
-	//
-	// TODO(acln): update this to mirror net.sysSocket completely:
```
mdlayher, line 467: After re-reading this just now, I should check out the upstream logic and probably just set this flag instead.

mdlayher: And fixed.

terinjokes (approved):

PR #185, opened 2021-02-20
implement signed integer encoding/decoding

### LINKED ISSUE #184: add support for signed integers encoding/decoding
Netlink has support for [signed integer attributes](https://elixir.bootlin.com/linux/v5.10.15/source/include/net/netlink.h#L104), which the library currently is missing. These are for example required when talking to rtnetlink to [set and modify the file descriptors of XDP programs](https://elixir.bootlin.com/linux/v5.10.15/source/net/core/rtnetlink.c#L2911) loaded to the links.

Providing these through the library would offer tested and validated methods of encoding these to the users, blocking workarounds through `Bytes` and `Uint32` implementations that could lead to bugs.

I've done some minor work on this on a [fork](https://github.com/fbegyn/netlink/tree/f-signed-ints), which is sufficient for my application, but can expand and modify it if to be upstreamed if this deemed as a desired feature.

fbegyn:
Resolves #184 by implementing the signed integer encoders and decoders.

The `int(uint(` is not necessarily pretty, but the other options to my knowledge are:
* unsafe pointer casting (like in `nlenc`)
* manually parsing the bytes (but then endianess is a thing)


mdlayher:
Thanks for this! I will try to take a look later today.

attribute.go
```diff
@@ -501,6 +561,63 @@ func (ae *AttributeEncoder) Uint64(typ uint16, v uint64) {
...
+}
+
+// Int16 encodes int16 data into an Attribute specified by typ.
+func (ae *AttributeEncoder) Int16(typ uint16, v int16) {
+	if ae.err != nil {
+		return
+	}
+
+	b := make([]byte, 2)
+	ae.ByteOrder.PutUint16(b,uint16(v))
```
mdlayher, line 583: I don't think this is gofmt'd.

attribute_test.go
```diff
@@ -909,18 +983,18 @@ func TestAttributeEncoderOK(t *testing.T) {
 			},
 		},
 		{
-			name:  "uint native endian",
+			name:  "uint-int32 native endian",
```
mdlayher, line 986: Just "int" for all these cases, not int32.

mdlayher (changes requested):
Overall seems fine. I would prefer not to add any new API to nlenc though: it is effectively deprecated since the addition of the attribute encoder/decoder types. Maybe I should officially add a deprecated tag on the package comment.

fbegyn:
With the `nlenc` being deprecated, I stripped out most of the code and tests in there, I only let the `put signed 32` test in there since that function is already part of the module. I've also ran `gofmt` against all the files again, looks like my last save didn't autocall it for some reason.

mdlayher (approved):
LGTM!

PR #187, opened 2021-08-18
netlink: add PID for netlink dial config

lwintermelon:
If you want to receive unicast message from kernel, you need to specify the PID of SockaddrNetlink when bind.
For example, you have a kernel module and you want to commuinicate with your userspace program.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -558,4 +558,8 @@ type Config struct {
 
 	// DisableNSLockThread is deprecated and has no effect.
 	DisableNSLockThread bool
+
+	// the unicast address of netlink socket. If set it to 0,
+	// the kernel takes care of assigning it.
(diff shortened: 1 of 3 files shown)


mdlayher:
Do you have any info about which kernel module(s) operate in this way? It's nothing I've ever run into. Thanks.

lwintermelon:
I mean if you make a kernel module that creates your own netlink protocol, and don't use generic netlink for some reasons.

mdlayher:
Thank you again for the PR, I've credited you for reporting this and have added the option in a new commit with an associated test: https://github.com/mdlayher/netlink/commit/53a1c10065e51077659ceedf921c8f0807abe8c0.

PR #195, opened 2022-01-11
Split tests that require circular dependencies

bdrung:
https://github.com/jsimonetti/rtnetlink and https://github.com/mdlayher/ethtool depend on netlink and therefore introduce a circular dependency. This library was packaged for Debian and circular dependencies should be avoided there (e.g. to ease
bootstrapping).

`rtnetlink` and `ethtool` are only needed for integration testing. So move all tests that require `rtnetlink` or `ethtool` into `conn_linux_consumer_test.go` and allow skipping running those test by specifying the build tag `break_circular_dependencies`.


mdlayher (commented):
Thanks. This has been on my radar and I'm not fundamentally opposed to the change, but I'm unsure what the current best practice is around this sort of thing. I wonder if it might make sense to have a separate internal go.mod for integration tests so that these can be avoided all together without needing build tags.

I'm AFK for a few days but will keep thinking on it.

bdrung:
Another alternative: Move these integration tests to the corresponding projects. If you have these integration tests in the netlink module to ensure that you don't break the API, you should add API test cases instead to netlink.

mdlayher:
Superseded by https://github.com/mdlayher/netlink/commit/a68a725f9dad2e01119be201a27a1dc26ba1fc59, will create a new release fairly soon.

PR #196, opened 2022-01-15
tests: Fix integer overflow on 32-bit architectures

bdrung:
On 32-bit architectures like armhf and i386, `math.MaxUint32` is bigger than the maximum value the signed 32-bit `int` can take. This causes the tests to fail to build:

```
conn_linux_integration_test.go:604:24: constant 4294967295 overflows int
FAIL	github.com/mdlayher/netlink [build failed]
```

So use `(*Rand) Uint32` instead which returns a pseudo-random 32-bit value as a `uint32`.

### DIFF
--- a/conn_linux_integration_test.go
+++ b/conn_linux_integration_test.go
@@ -6,7 +6,6 @@ package netlink_test
 import (
 	"errors"
 	"fmt"
-	"math"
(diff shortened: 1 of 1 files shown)


mdlayher:
Thanks, good catch!

Issue #179, opened 2021-01-24
netlink: consider memoizing the largest seen buffer size in Conn.Receive


mdlayher:
Right now we allocate a page of memory, peek into the socket, and double the size of the buffer if we can't fit the entire message.

For APIs which frequently return a large amount of data, it probably makes sense to memoize the largest used buffer on a per-Conn basis so that we know how much to allocate up front rather than looping with peek/allocate.

mdlayher:
Before committing to doing this with a public API, I may do something along the lines of `NLDEBUG=experiments=conn-memoize-receive` or similar so this can be quickly toggled with no code changes in deployments.

helios741:
I used netlink for internal services and optimized [conn.Receive](https://github.com/mdlayher/netlink/blob/main/conn_linux.go#L120) with sync.Pool, saving GCMark time and 28% CPU savings.
If possible, I can submit a PR to optimize conn.Receive.

Before optimisation:
CPU:
![image](https://user-images.githubusercontent.com/12036324/165299982-2ce16669-5bc5-4a9c-a923-207aa7b6b4d9.png)

mem:
![image](https://user-images.githubusercontent.com/12036324/165300047-e4732e9b-0508-4e94-bc1d-0beb3224b002.png)

After optimization:
CPU:
![image](https://user-images.githubusercontent.com/12036324/165300164-f1814a2f-f2e3-4bbb-96a6-08055ae01e55.png)

Mem:
![image](https://user-images.githubusercontent.com/12036324/165300269-1aa1dfc5-3897-48e3-ab70-0369393ddf48.png)

mdlayher:
I don't believe a sync.Pool optimization would be safe because the library hands out allocated buffers for caller use, and putting those back in the pool could mean they are clobbered.

I think a more appropriate approach to reuse memory would be to add an API which reads into a caller-allocated buffer and leaves managing that up to the caller.

Issue #178, opened 2021-01-24
netlink: add a version of Conn.Receive that doesn't allocate its own buffers


mdlayher:
This could speed things up significantly for big payloads, and then we can also get rid of the peek/loop/allocate logic and rely on the caller to provide a big enough buffer to read out all the available messages.

nickgarlis:
I've opened https://github.com/mdlayher/netlink/pull/270 as an alternative. @mdlayher Let me know if this is not aligned with what you had in mind. I'll let the PR simmer for a few days but will probably merge if there are no objections.

PR #200, opened 2022-09-30
Dynamic socket buffer allocation

turekt:
Hi,

the motivation behind this PR has several elements:
- I have noticed that when nftables returns a big response, the socket read can return ENOBUFS (no buffer space available), this was explained and discussed here: https://github.com/google/nftables/pull/191/files/0d4369aacbd8b10bc86765a69851d0d01a821fd8#r979436375
- I tried to play around with a fix for this in the netlink library in order to try and receive all of the data after ENOBUFS is returned but this is just not possible (from netlink man page):
```
Netlink is not a reliable protocol.  It tries its best to deliver a message to its destination(s), but may drop messages when an out-of-memory condition or other error occurs.  For reliable transfer the sender can  request  an
acknowledgement  from  the  receiver  by setting the NLM_F_ACK flag.  An acknowledgment is an NLMSG_ERROR packet with the error field set to 0.  The application must generate acknowledgements for received messages itself.  The
kernel tries to send an NLMSG_ERROR message for every failed packet.  A user process should follow this convention too.

However, reliable transmissions from kernel to user are impossible in any case.  The kernel can't send a netlink message if the socket buffer is full: the message will be dropped and the kernel and the user-space process  will
no longer have the same view of kernel state.  It is up to the application to detect when this happens (via the ENOBUFS error returned by recvmsg(2)) and resynchronize.
```

- If ENOBUFS happen, part of the message will be dropped for sure. Therefore, it is not possible to recover from this error but to re-initiate the same message again with a bigger buffer. Additionally, I saw issues #178 and #179 and it seemed to me that both changes can be covered with this PR.

With this in mind, since it was discussed to try and address the ENOBUFS issue in the netlink lib (https://github.com/google/nftables/pull/191/files/0d4369aacbd8b10bc86765a69851d0d01a821fd8#r982856106), I am introducing a PR for discussion which covers:
- introduces a `ReceiveBuffer` (`ExecuteBuffer`) func which receives a `BufferAllocationFunc` that allocates buffers for the underlying socket, passed by the user (covering issue #178)
- introduces a default `BufferAllocationFunc` (or default allocation strategy) which is similar to the previous peek-loop-allocate but in case ENOBUFS happens, it automatically resizes the socket read and write buffers
- exposed the `ReadBuffer` and `WriteBuffer` methods via `bufferGetter` interface for easier calculation of buffer size by user applications
- already implemented functions were minimally changed to retain backwards compatibility

In the end, I am not sure if this is the best approach since applications could still catch the ENOBUFS error themselves, resize the read and write buffers with `SetReadBuffer` or `SetWriteBuffer` and then resend the message making this PR unnecessary? We can change the PR as per your feedback.

Let me know what you think.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -10,6 +10,8 @@ import (
 	"golang.org/x/net/bpf"
 )
 
+type BufferAllocationFunc func() ([]byte, error)
+
 // A Conn is a connection to netlink.  A Conn can be used to send and
 // receives messages to and from netlink.
 //
@@ -53,6 +55,7 @@ type Socket interface {
 	Send(m Message) error
 	SendMessages(m []Message) error
 	Receive() ([]Message, error)
+	ReceiveBuffer(fn BufferAllocationFunc) ([]Message, error)
 }
 
 // Dial dials a connection to netlink, using the specified netlink family.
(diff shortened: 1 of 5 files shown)


nickgarlis:
Hi @turekt

There have been several changes in both github.com/google/nftables and this package to address these issues. In particular, the copy buffer is now configurable and resizing the socket buffer is now being done in nftables.

Configuration like this tends to be specific to each Netlink family, so it’s likely better handled by consumers rather than in this library.

I’m going to close this PR for now, but feel free to reopen or submit an updated version if you think there’s a better approach.

PR #210, opened 2024-03-12
feat: Handle OobData

dave-tucker:
Adds OobData []byte to the Message struct.
During the Recieve() function, this field is populated with the raw bytes of any control messages recieved from the recvmsg call.

An example of how to process the OobData field has been included.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -590,4 +590,10 @@ type Config struct {
 	// When possible, setting Strict to true is recommended for applications
 	// running on modern Linux kernels.
 	Strict bool
+
+	// EnableControlMessages enables the use of control messages for
+	// netlink. This option is intended for advanced use cases where the
+	// caller needs to receive control messages that contain ancillary data.
+	// For example, when using options like `ListenAllNSID`.
+	EnableControlMessages bool
 }
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -19,7 +19,8 @@ var _ Socket = &conn{}
 
 // A conn is the Linux implementation of a netlink sockets connection.
 type conn struct {
-	s *socket.Conn
+	s                     *socket.Conn
+	enableControlMessages bool
 }
 
 // dial is the entry point for Dial. dial opens a netlink socket using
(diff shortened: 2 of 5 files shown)


conn_linux.go
```diff
@@ -141,12 +139,23 @@ func (c *conn) Receive() ([]Message, error) {
...
+	}
+
 	// Read out all available messages
-	n, _, _, _, err := c.s.Recvmsg(context.Background(), b, nil, 0)
+	n, oobn, _, _, err := c.s.Recvmsg(context.Background(), b, oob, 0)
 	if err != nil {
 		return nil, err
 	}
 
+	var rawOob []byte
```
mdlayher, line 154:
```suggestion
	var rawOOB []byte
```

conn.go
```diff
@@ -590,4 +590,10 @@ type Config struct {
 	// When possible, setting Strict to true is recommended for applications
 	// running on modern Linux kernels.
 	Strict bool
+
+	// EnableControlMessages enables the use of control messages for
+	// netlink. This option is intended for advanced use cases where the
+	// caller needs to receive control messages that contain ancillary data.
+	// For example, when using options like `ListenAllNSID`.
```
mdlayher, line 597: Add another sentence like:

```
// When enabled, the Message.OOBData field will be populated with control message data.
```

example_test.go
```diff
@@ -108,3 +109,59 @@ func exampleAttributes() []byte {
...
+		Groups: rtmGroupLink,
+		// Enable control messages to receive the netnsid
+		// since we're going to set NETLINK_LISTEN_ALL_NSID
+		EnableControlMessages: true,
+	})
+	if err != nil {
+		log.Fatalf("failed to dial netlink: %v", err)
+	}
+	defer c.Close()
+	c.SetOption(netlink.ListenAllNSID, true)
```
mdlayher, line 135: Check error.

example_test.go
```diff
@@ -108,3 +109,59 @@ func exampleAttributes() []byte {
...
+		// Listen for netlink messages triggered by multicast groups
+		msgs, err := c.Receive()
+		if err != nil {
+			log.Fatalf("failed to receive messages: %v", err)
+		}
+
+		// Iterate over recieved messages and print them
+		for _, msg := range msgs {
+			// If the message contains oob data, parse it and print the netnsid
+			if msg.OobData != nil {
```
mdlayher, line 147: Invert conditional and continue if nil rather than adding indentation.

PR #219, opened 2025-01-04
debug: option to print using mnl format

aojea:
nftables uses the libmnl library that has a particular debugging format. This format is also used for nftables --debug=all. google/nftables golang library to interact with nftables uses this library to interface with netlink directly.
Since these libraries are not well documented, is common to have to dump the netlink bits to compare.

This commit implements an option to choose the same format used in nftables by adding a new option to the NLDEBUG environmental variable and using the format=mnt key value.

Output of `nft --debug=all`

image

Output of a `nftables` dump using google/nftables and setting the environment variable
https://gist.github.com/aojea/8a811e11a060fa638aa2dd8165960d5a

`NLDEBUG=format=mnl go run main.go cni-kindnet prerouting`

image

cc: @stapelberg


debug.go
```diff
@@ -60,10 +69,200 @@ func newDebugger(args []string) *debugger {
 // high enough to print the message.
 func (d *debugger) debugf(level int, format string, v ...interface{}) {
 	if d.Level >= level {
-		d.Log.Printf(format, v...)
+		if d.Format == "mnl" {
+			for _, iface := range v {
+				if msg, ok := iface.(Message); ok {
+					nlmsgFprintf(d.Log.Writer(), msg)
+				}
```
stapelberg, line 76: Shouldn’t we call `d.Log.Printf()` in the else branch here? If not, can you add a comment explaining why this code path only handles values of type `Message`?

doc.go
```diff
@@ -27,7 +27,10 @@
 //
 //	$ NLDEBUG=level=1 ./nlctl
 //
+//	$ NLDEBUG=level=1,format=mnl ./nlctl
+//
 // Available key/value debugger options include:
 //
 //	level=N: specify the debugging level (only "1" is currently supported)
+//	format=mnl: specify the same format used by libmnl (nftables --debug=all)
```
stapelberg, line 35: nit: `nft --debug=all`, not `nftables`

stapelberg (commented):
Thanks for your PR!

debug.go
aojea: oh, it was an oversight,

aojea: I refactored the code to remove one indentation layer and added a switch statement if there is need to add different output formats, I can see pyroute2 format is also used in different places

mdlayher (commented):
This is awesome, I would be fine making this the default debugging format given that the existing one is probably not very useful for most circumstances.

I do think that the printer is complex enough to probably at least merit a unit test that dumps a fixed output and verifies that it does not drift.

aojea:
> This is awesome, I would be fine making this the default debugging format given that the existing one is probably not very useful for most circumstances.
> 
> I do think that the printer is complex enough to probably at least merit a unit test that dumps a fixed output and verifies that it does not drift.

@mdlayher added a commit to bump github actions and golang versions and defaulted the logger to `mnl`

aojea:
@mdlayher my apologies, somehow I didn't realize the PR was not updated until today, it should be ready for review now

Issue #227, opened 2025-08-26
Unit tests fail en masse on s390x (and presumably other bigendian archs)


dswarbrick:
Since v1.8.0, many unit tests fail due to what appear to be endian decoding issues:
```
$ go version
go version go1.24.4 linux/s390x

$ go test -skip TestIntegration ./...
--- FAIL: TestMarshalAttributes (0.00s)
    --- FAIL: TestMarshalAttributes/one_attribute,_no_data (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x04 0x00 0x01 0x00]
            -  got: [0x00 0x04 0x00 0x01]
    --- FAIL: TestMarshalAttributes/one_attribute,_no_data,_length_calculated (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x04 0x00 0x01 0x00]
            -  got: [0x00 0x04 0x00 0x01]
    --- FAIL: TestMarshalAttributes/one_attribute,_padded (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x05 0x00 0x01 0x00 0xff 0x00 0x00 0x00]
            -  got: [0x00 0x05 0x00 0x01 0xff 0x00 0x00 0x00]
    --- FAIL: TestMarshalAttributes/one_attribute,_padded,_length_calculated (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x05 0x00 0x01 0x00 0xff 0x00 0x00 0x00]
            -  got: [0x00 0x05 0x00 0x01 0xff 0x00 0x00 0x00]
    --- FAIL: TestMarshalAttributes/one_attribute,_aligned (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x08 0x00 0x02 0x00 0xaa 0xbb 0xcc 0xdd]
            -  got: [0x00 0x08 0x00 0x02 0xaa 0xbb 0xcc 0xdd]
    --- FAIL: TestMarshalAttributes/one_attribute,_aligned,_length_calculated (0.00s)
        attribute_test.go:227: unexpected bytes:
            - want: [0x08 0x00 0x02 0x00 0xaa 0xbb 0xcc 0xdd]
            -  got: [0x00 0x08 0x00 0x02 0xaa 0xbb 0xcc 0xdd]
    --- FAIL: TestMarshalAttributes/multiple_attributes (0.00s)
        attribute_test.go:227: unexpected bytes:
[...]

PR #220, opened 2025-01-13
use Go 1.21's binary.NativeEndian

bradfitz:

PR #276, opened 2026-04-06
Add bigendian test fixtures for nlenc package

dswarbrick:
Add distinct test fixtures for bigendian hosts to nlenc package and drop skipBigEndian function.

IMHO these tests are a little paranoid, testing and re-testing the same byte-swapping scenarios, just with different values. The number of tests cases could probably be reduced by half.

Also, I question the entire need for these functions, given that such functionality exists in the Go standard library (e.g. `binary.NativeEndian.Uint32` and friends).

Refs: #220, #227

### DIFF
--- a/nlenc/int.go
+++ b/nlenc/int.go
@@ -7,6 +7,8 @@ import (
 
 // PutUint8 encodes a uint8 into b.
 // If b is not exactly 1 byte in length, PutUint8 will panic.
+//
+// Deprecated: Use inline "b[n] = v" instead.
 func PutUint8(b []byte, v uint8) {
 	if l := len(b); l != 1 {
 		panic(fmt.Sprintf("PutUint8: unexpected byte slice length: %d", l))
@@ -65,6 +67,8 @@ func PutInt32(b []byte, v int32) {
 
 // Uint8 decodes a uint8 from b.
 // If b is not exactly 1 byte in length, Uint8 will panic.
+//
+// Deprecated: Use inline "b[n]" instead.
 func Uint8(b []byte) uint8 {
 	if l := len(b); l != 1 {
 		panic(fmt.Sprintf("Uint8: unexpected byte slice length: %d", l))
@@ -123,6 +127,8 @@ func Int32(b []byte) int32 {
 
 // Uint8Bytes encodes a uint8 into a newly-allocated byte slice. It is a
 // shortcut for allocating a new byte slice and filling it using PutUint8.
+//
+// Deprecated: Use inline "[]byte{v}" instead.
 func Uint8Bytes(v uint8) []byte {
 	return []byte{v}
 }
--- a/nlenc/int.go
+++ b/nlenc/int.go
@@ -1,8 +1,8 @@
 package nlenc
 
 import (
+	"encoding/binary"
 	"fmt"
-	"unsafe"
 )
 
 // PutUint8 encodes a uint8 into b.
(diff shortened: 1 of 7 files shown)


nickgarlis:
> Also, I question the entire need for these functions, given that such functionality exists in the Go standard library (e.g. binary.NativeEndian.Uint32 and friends).

I think I agree that some of these `nlenc` functions could be removed. My understanding is that they were there before the standard lib got its own implementation. I am therefore hesitant to remove them without releasing a new major version. I am not sure what the implications would be for people importing them still.

> IMHO these tests are a little paranoid, testing and re-testing the same byte-swapping scenarios, just with different values. The number of tests cases could probably be reduced by half.

Probably not worth investing too much time in it if they are to be removed in the future.

dswarbrick:
> I think I agree that some of these `nlenc` could be removed. My understanding is that they were there before the standard lib got its own implementation. I am therefore hesitant to remove them without releasing a new major version. I am not sure what the implications would be for people importing them still.

The `Uint32` / `PutUint32` etc functions have existed since Go 1.0. The only thing that is a bit more recent is `binary.NativeEndian` (added in Go 1.21), but this project previously used https://github.com/josharian/native to determine endianness, so in theory these functions were never really necessary.

Since the functions are exported and potentially being used by third-party packages, how about simply making them wrappers for the standard library functions? This will effectively eliminate the need to ship unit tests.

nickgarlis:
> Since the functions are exported and potentially being used by third-party packages, how about simply making them wrappers for the standard library functions? This will effectively eliminate the need to ship unit tests.

Yes I like this idea. We would probably still need one or two smoketests to make sure the abstraction works as intended but nothing as exhaustive as we currently have.

dswarbrick:
> Yes I like this idea. We would probably still need one or two smoketests to make sure the abstraction works as intended but nothing as exhaustive as we currently have.

I have made the existing int helper functions wrappers for their respective functions in `binary.NativeEndian`, and marked them deprecated in the function comments. This has in turn triggered golangci-lint warnings elsewhere in the code. Before I fix those, I would appreciate a thumbs-up if the general approach of this PR is acceptable.

I haven't deprecated the `PutUint8` / `Uint8`, but given how extremely trivial these functions are, I recommend adding a deprecation comment to these anyway. They should never have been functions - the overhead of calling such a function is going to outweigh doing a simple inline type conversion like `b := []byte{v}` or `v := b[0]`. If I didn't know that this code predated vibe coding, I would have guessed that it was AI slop.

nlenc/int.go
```diff
@@ -17,42 +17,50 @@ func PutUint8(b []byte, v uint8) {
 
 // PutUint16 encodes a uint16 into b using the host machine's native endianness.
 // If b is not exactly 2 bytes in length, PutUint16 will panic.
```
nickgarlis, line 21: The functions will no longer panic but I don't think that's a big deal

nickgarlis:
That's what I had in mind. Thanks!

nlenc/int.go
dswarbrick: The statement about panicking on invalid length is still true - unless you want me to remove the length check as well, making the functions _very_ thin wrappers around `binary.NativeEndian`.

The standard lib functions will panic if the slice is too short, but they will tolerate extra bytes in the slice.

nickgarlis: Sorry, I was going through this quickly and didn't notice that the length check was still there. It's good as is.

nickgarlis:
> I haven't deprecated the PutUint8 / Uint8, but given how extremely trivial these functions are, I recommend adding a deprecation comment to these anyway. 

I agree, no point in supporting them when the rest are deprecated anyways

dswarbrick:
All done here. Ready to merge.

nickgarlis (approved):

PR #280, opened 2026-04-10
conn: align receive buffer to nlmsgAlign boundary

### LINKED ISSUE #279: panic: runtime error: slice bounds out of range
Seen after Bump github.com/mdlayher/netlink from 1.8.0 to 1.10.0 in https://github.com/kubernetes-sigs/kindnet/pull/71

Issue https://github.com/kubernetes-sigs/kindnet/issues/72

aojea:
getBuffer() used MSG_PEEK|MSG_TRUNC to determine the exact byte count of the next message, then allocated a buffer of exactly that size. When the peeked length is not a multiple of 4 (the netlink alignment boundary), parseMessagesIter rounds the slice index up with nlmsgAlign, which immediately panics with "slice bounds out of range".

Fixes: https://github.com/mdlayher/netlink/issues/279

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -147,7 +147,8 @@ func (c *conn) Receive() ([]Message, error) {
 // getBuffer returns the buffer to use for receiving messages and a function to
 // release it back to the pool if applicable. If the pool is not configured, a
 // new buffer is allocated by peeking the size of the next message to be
-// received.
+// received. The buffer size is aligned to the next multiple of the alignment of a netlink
+// message, to avoid panic when parsing messages from the buffer.
 func (c *conn) getBuffer() ([]byte, func(), error) {
 	if c.pool != nil {
 		bp := c.pool.Get().(*[]byte)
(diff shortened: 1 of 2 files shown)


nickgarlis:
This looks good. Thanks @aojea !

Question: The main branch is now on Go 1.25. Would you prefer a patch on 1.24 or are you okay with upgrading ?

nickgarlis (approved):

aojea:
> This looks good. Thanks @aojea !
> 
> Question: The main branch is now on Go 1.25. Would you prefer a patch on 1.24 or are you okay with upgrading ?

I didn;t find any issue so far on my projects, so this lgtm

aojea:
@nickgarlis can you cut a new release? 

I wonder if we should use the retract directive to mark the v1.10.0 version https://go.dev/ref/mod#go-mod-file-retract

nickgarlis:
> @nickgarlis can you cut a new release?

@aojea I can do that. I was initially thinking of doing `v1.10.1` based on [this branch](https://github.com/mdlayher/netlink/tree/release-1.10) to keep the release in Go 1.24. But since we are okay with 1.25, I'll do a `v1.11.0` instead. Let me know if you have any objections.

> I wonder if we should use the retract directive to mark the v1.10.0 version https://go.dev/ref/mod#go-mod-file-retract

Yes we should. Thanks for the suggestion.

PR #284, opened 2026-04-13
conn_linux: copy pooled receive buffers before parsing

ivanlevitsky:
## Problem
On Linux, when `Config.MessageBufferSize` is enabled, receive buffers are reused from a pool.
`Message.UnmarshalBinary` keeps `Message.Data` as a slice into the parsed input buffer, so returned messages may reference pooled memory that is reused by later receives.

This is especially visible when responses are read across multiple `recvmsg` calls (for example, multipart netlink dumps): a later receive can overwrite data from previously returned messages.

## Root cause
`ReceiveIter` parsed messages directly from the pooled receive buffer and then returned that buffer to the pool before callers were necessarily done using `Message.Data`.

## Fix
- Add `ownedBuffer` in `conn_linux.go` and parse from it in `ReceiveIter`.
- Non-pooled path: keep existing behavior (`b[:nlmsgAlign(n)]`) with no extra copy.
- Pooled path: copy `n` received bytes into a fresh aligned buffer and parse from that owned memory.

This keeps the change minimal and targeted to the pooled-buffer path only.

## Tests
Add integration regression test:
- `TestIntegrationConnMessageBufferSizeCopiesMessageData` in `conn_linux_integration_test.go`.
- Uses `NETLINK_USERSOCK` with `MessageBufferSize` enabled.
- Sends two datagrams with different payloads.
- Verifies payload from the first `Receive()` remains unchanged after the second `Receive()`.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -163,6 +163,21 @@ func (c *conn) getBuffer() ([]byte, func(), error) {
 	return make([]byte, nlmsgAlign(n)), func() {}, nil
 }
 
+// ownedBuffer trims b to the aligned size of the received datagram and, when
+// using pooled buffers, copies it to memory owned by the caller.
+func (c *conn) ownedBuffer(b []byte, n int) []byte {
+	aligned := nlmsgAlign(n)
+	if c.pool == nil {
+		return b[:aligned]
+	}
+
+	// A pooled buffer may not have an aligned length. Copy the received bytes
+	// to an aligned buffer and let the zero-value tail bytes act as padding.
+	out := make([]byte, aligned)
+	copy(out, b[:n])
+	return out
+}
+
 // ReceiveIter returns an iterator over Messages received from netlink.
 func (c *conn) ReceiveIter() iter.Seq2[Message, error] {
 	return func(yield func(Message, error) bool) {
@@ -190,7 +205,7 @@ func (c *conn) ReceiveIter() iter.Seq2[Message, error] {
 			return
 		}
 
-		for msg, err := range parseMessagesIter(b[:nlmsgAlign(n)]) {
+		for msg, err := range parseMessagesIter(c.ownedBuffer(b, n)) {
 			if err != nil {
 				yield(Message{}, err)
 				return
--- a/conn_linux_integration_test.go
+++ b/conn_linux_integration_test.go
@@ -1158,6 +1158,94 @@ func TestIntegrationConnMessageBufferSize(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


nickgarlis:
Thanks for this PR @ivanlevitsky !

I had a quick look and I think this looks good. I am wondering whether another alternative could be to copy into `Message.Data` during `Message.UnmarshalBinary`. This could maybe simplify things and copy less bytes overall. Of course, this would have the downside of slightly penalizing performance for `Message.UnmarshalBinary` but I don't think that's not significant enough. WDYT ?

ivanlevitsky:
> Thanks for this PR @ivanlevitsky !
> 
> I had a quick look and I think this looks good. I am wondering whether another alternative could be to copy into `Message.Data` during `Message.UnmarshalBinary`. This could maybe simplify things and copy less bytes overall. Of course, this would have the downside of slightly penalizing performance for `Message.UnmarshalBinary` but I don't think that's not significant enough. WDYT ?

Thank you, this is a very reasonable alternative.

I did consider copying in `Message.UnmarshalBinary`, but I intentionally kept the copy out of that layer:

- `UnmarshalBinary` is currently a low-level zero-copy primitive used in multiple paths. Copying there would introduce allocations globally, including paths not affected by this bug.
- This issue is specific to Linux pooled receives (`MessageBufferSize`), so fixing it in `conn_linux` keeps the overhead localized to the problematic path.
- The current change does one allocation/copy per `recvmsg` datagram in the pooled path. Copying in `UnmarshalBinary` would typically allocate per parsed message, which is usually less favorable for multipart responses.

So I’d prefer to keep this targeted fix. If you’d like, we can discuss changing `UnmarshalBinary` ownership semantics separately.

nickgarlis:
Thanks for the detailed explanation!

> If you’d like, we can discuss changing UnmarshalBinary ownership semantics separately.

Yes, do let me know if you think there are ways it can be improved.

nickgarlis (approved):

conn_linux.go
```diff
@@ -190,7 +205,7 @@ func (c *conn) ReceiveIter() iter.Seq2[Message, error] {
 			return
 		}
 
-		for msg, err := range parseMessagesIter(b[:nlmsgAlign(n)]) {
+		for msg, err := range parseMessagesIter(c.ownedBuffer(b, n)) {
```
nickgarlis, line 208: ~Last question; This will also copy when `MessageBufferSize` is not set. Was that intentional in favor of simplicity ?~

Issue #286, opened 2026-04-17
Idea: struct tag-based attribute codec


nickgarlis:
Not sure if this belongs here or as a separate package, but I've been experimenting with a reflection-based codec that maps struct tags to netlink attributes and wanted to see if there's any interest.

## The idea

Instead of writing manual encode/decode per struct like this:

```go
// Decoding
ad.ByteOrder = binary.BigEndian
for ad.Next() {
    switch ad.Type() {
    case unix.NFTA_TABLE_NAME:
        t.Name = ad.String()
    case unix.NFTA_TABLE_USE:
        t.Use = ad.Uint32()
    case unix.NFTA_TABLE_FLAGS:
        t.Flags = TableFlags(ad.Uint32())
    }
}

// Encoding
ae := netlink.NewAttributeEncoder()
ae.ByteOrder = binary.BigEndian
ae.String(unix.NFTA_TABLE_NAME, t.Name)
ae.Uint32(unix.NFTA_TABLE_FLAGS, uint32(t.Flags))
data, err := ae.Encode()
```

You'd just tag the struct:

```go
type Table struct {
    Name  string  `netlink:"1,be"`
    Use   *uint32 `netlink:"4,be"`
    Flags *uint32 `netlink:"6,be"`
}

tbl := Table{Name: "filter", Flags: ptr(0x1)}

codec := netlink.NewAttributeCodec()
attrs, err := codec.Marshal(tbl)
err := codec.Unmarshal(b, &tbl)
```

The codec figures out the encoding from the Go type:

- `uint*`: scalar (native endian, or `be`/`le` tag)
- `string`: NUL-terminated
- `[]byte`: binary blob
- `bool`: flag (presence = true)
- `struct`: nested attributes
- `*T`: optional (nil = absent)
- `[]T` with `multi`: repeated attribute
- `[]T` with `indexed`: indexed array (children keyed 0, 1, 2…)

## Limitations

Uses reflection, so probably slower than hand-written code, though I haven't benchmarked it yet. The struct metadata is cached so at least that part is paid once.

## Implementation

I have a working implementation (~600 lines) with some test coverage. The tag style and spec is inspired by [YNL](https://docs.kernel.org/userspace-api/netlink/specs.html) which gives a rough idea of a spec that applies to most families. I'll open a draft PR and leave it up in case anyone wants to take a look or give feedback.

nickgarlis:
[...]

PR #287, opened 2026-04-17
netlink: add struct tag-based attribute codec

nickgarlis:
Add a reflection-based Codec that maps struct fields tagged with `netlink:"<index>[,opts]"` to netlink attributes, providing a declarative alternative to manual AttributeEncoder/AttributeDecoder usage.

Supported types: scalars (uint*/int* with optional be/le byte order), strings, []byte, bools (flags), nested structs, pointers (optional), slices (multi/indexed layouts), and sub-messages via resolver functions.

Struct metadata is cached per type to avoid repeated reflection overhead.

See #286.

### DIFF
--- a/attribute_codec.go
+++ b/attribute_codec.go
@@ -1,0 +1,707 @@
+package netlink
+
+import (
+	"encoding/binary"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/mdlayher/netlink/nlenc"
+)
+
+type SubMessageResolver func(selector any) (any, error)
+
+// AttributeCodecConfig holds optional configuration for a Codec.
+type AttributeCodecConfig struct {
+	// Resolvers map sub-message tag names to functions that return
+	// a concrete struct pointer for a given selector value.
+	Resolvers map[string]SubMessageResolver
+}
+
+// Codec is a stateful encoder/decoder that holds configuration.
+type Codec struct {
+	resolvers map[string]SubMessageResolver
+}
+
+// NewAttributeCodec creates a new Codec with the given configuration.
+// Pass nil for families that need no special configuration.
+func NewAttributeCodec(cfg *AttributeCodecConfig) *Codec {
+	c := &Codec{}
+	if cfg != nil {
+		c.resolvers = cfg.Resolvers
+	}
+	return c
+}
+
+// Marshal encodes a struct into netlink attribute bytes.
+func (c *Codec) Marshal(v any) ([]byte, error) {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() == reflect.Pointer {
+		rv = rv.Elem()
+	}
+	if rv.Kind() != reflect.Struct {
+		return nil, fmt.Errorf("netlink: Marshal requires a struct or pointer to struct, got %T", v)
+	}
+	attrs, err := c.encodeAttrs(rv)
+	if err != nil {
+		return nil, err
+	}
+	return MarshalAttributes(attrs)
+}
+
+// Unmarshal decodes netlink attribute bytes into a struct.
+func (c *Codec) Unmarshal(b []byte, v any) error {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() != reflect.Pointer || rv.IsNil() {
+		return fmt.Errorf("netlink: Unmarshal requires a non-nil pointer to struct, got %T", v)
+	}
+	rv = rv.Elem()
+	if rv.Kind() != reflect.Struct {
+		return fmt.Errorf("netlink: Unmarshal requires a pointer to struct, got %T", v)
+	}
+	attrs, err := UnmarshalAttributes(b)
+	if err != nil {
+		return err
+	}
+	return c.decodeAttrs(attrs, rv)
+}
+
+type attrType uint8
+
+const (
+	_ attrType = iota
+	attrScalar
+	attrString
+	attrBinary
+	attrFlag
+	attrNest
+)
+
+type attrLayout uint8
+
+const (
+	layoutSingle  attrLayout = iota // single value
+	layoutMulti                     // repeated attribute (same index)
+	layoutIndexed                   // nested with sequential uint16 indices
+)
+
+type fieldTag struct {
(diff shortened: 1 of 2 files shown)


nickgarlis:
Opinions are welcome. I am not sure whether it's a good idea to add another attribute encoder to the package but it could simplify things a lot by reducing boilerplate.

Tagging a few people here who might be interested: @aojea @SuperQ @florianl

aojea:
what will be the use cases for this? can you provide some realistic ones?

nickgarlis:
> what will be the use cases for this? can you provide some realistic ones?

Taking nftables as an example, for each attribute set you end up manually writing serialization and deserialization, which adds up to a lot of boilerplate. A typical table encode/decode looks something like this:

```golang
func tableFromAttrs(ad *netlink.AttributeDecoder) Table {
    var t Table
    ad.ByteOrder = binary.BigEndian
    for ad.Next() {
        switch ad.Type() {
        case unix.NFTA_TABLE_NAME:
            t.Name = ad.String()
        case unix.NFTA_TABLE_FLAGS:
            t.Flags = ad.Uint32()
        case unix.NFTA_TABLE_USE:
            v := ad.Uint32()
            t.Use = &v
        case unix.NFTA_TABLE_HANDLE:
            v := ad.Uint64()
            t.Handle = &v
        }
    }
    return t
}

func tableToAttrs(t *Table) ([]byte, error) {
    ae := netlink.NewAttributeEncoder()
    ae.ByteOrder = binary.BigEndian
    ae.String(unix.NFTA_TABLE_NAME, t.Name)
    ae.Uint32(unix.NFTA_TABLE_FLAGS, t.Flags)
    if t.Use != nil {
        ae.Uint32(unix.NFTA_TABLE_USE, *t.Use)
    }
    if t.Handle != nil {
        ae.Uint64(unix.NFTA_TABLE_HANDLE, *t.Handle)
    }
    return ae.Encode()
}
```

With a tag-based codec, the struct definition becomes the single source of truth  similar to how encoding/json or gopkg.in/yaml.v3 work:

```golang
type Table struct {
    Name   string  `netlink:"1,be"`  // NFTA_TABLE_NAME
    Flags  uint32  `netlink:"2,be"`  // NFTA_TABLE_FLAGS
    Use    *uint32 `netlink:"3,be"`  // NFTA_TABLE_USE
    Handle *uint64 `netlink:"4,be"`  // NFTA_TABLE_HANDLE
}

codec := netlink.NewAttributeCodec(nil)

// encode
b, err := codec.Marshal(t)

// decode
var t Table
err := codec.Unmarshal(b, &t)
```

I started exploring this while building a YNL code generator for Go. In YNL, operations often use subsets of a shared attribute set. For example, newtable, gettable, deltable, and destroytable all reference table-attrs but each uses different fields. A naive generator would emit a separate struct (plus encode/decode boilerplate) for every operation/direction combination, which quickly explodes in families like nftables with dozens of operations. With a tag-based codec, the generator can instead emit one struct per attribute set using pointer fields for optional attributes, and get serialization for free.

nickgarlis:
Here’s an example of a more advanced case where discrimination is needed. The tag options `selector` and `sub-msg` follow the YNL spec: https://elixir.bootlin.com/linux/v7.0/source/Documentation/netlink/specs/nftables.yaml#L433 

```golang
type Expression interface {
    exprType() string
}

type ExprMeta struct {
    Dreg *uint32 `netlink:"1,be"`
    Key  uint32  `netlink:"2,be"`
    Sreg *uint32 `netlink:"3,be"`
}

func (*ExprMeta) exprType() string { return "meta" }

type ExprCounter struct {
    Bytes   uint64 `netlink:"1,be"`
    Packets uint64 `netlink:"2,be"`
}

func (*ExprCounter) exprType() string { return "counter" }

// "Data" is resolved based the "Name" field.
type ExprAttrs struct {
    Name string     `netlink:"1,be"`
    Data Expression `netlink:"2,submsg=expr-ops,selector=Name"`
}

type RuleAttrs struct {
    Table string      `netlink:"1,be"`
    Chain string      `netlink:"2,be"`
    Exprs []ExprAttrs `netlink:"4,indexed"`
}

codec := netlink.NewAttributeCodec(&netlink.AttributeCodecConfig{
    Resolvers: map[string]netlink.SubMessageResolver{
        "expr-ops": func(selector any) (any, error) {
            switch selector.(string) {
            case "meta":
                return &ExprMeta{}, nil
            case "counter":
                return &ExprCounter{}, nil
            default:
                return nil, fmt.Errorf("unknown expression: %s", selector)
            }
        },
    },
})

rule := RuleAttrs{
    Table: "filter",
    Chain: "input",
    Exprs: []ExprAttrs{
        {Name: "meta", Data: &ExprMeta{Key: 16, Dreg: ptr(uint32(1))}},
        // "cmp" expression omitted for brevity
        {Name: "counter", Data: &ExprCounter{}},
    },
}

b, err := codec.Marshal(rule)

var got RuleAttrs
err = codec.Unmarshal(b, &got)
// got.Exprs[0].Data.(*ExprMeta).Key == 16
// got.Exprs[1].Data.(*ExprCounter) != nil
```

attribute_codec.go
```diff
@@ -0,0 +1,707 @@
...
+	attrs, err := c.encodeAttrs(rv)
+	if err != nil {
+		return nil, err
+	}
+	return MarshalAttributes(attrs)
+}
+
+// Unmarshal decodes netlink attribute bytes into a struct.
+func (c *Codec) Unmarshal(b []byte, v any) error {
+	rv := reflect.ValueOf(v)
```
florianl, line 56: I generally avoid functions like reflect.ValueOf() because they hinder the Go compiler's ability to perform dead code elimination. This can lead to a significant and unnecessary increase in the final binary size. While hand-writing logic for known structs is more verbose, the resulting performance and footprint optimizations are well worth the manual effort.

nickgarlis: Thanks for having a look! Yeah I am not too keen on it being slower. Admittedly, I hadn't considered the binary size consequences. The idea was that it might be useful for use cases where the performance tradeoffs are acceptable. If there are no such use cases then having and maintaining this feature is probably not worth it.

PR #291, opened 2026-05-02
conn: fix Receive blocking Send

### LINKED ISSUE #288: Can no longer handle NFQUEUE reads from different goroutine than sending verdicts in v1.11
We use this library via florianl/go-nfqueue and we have one goroutine reading NFQUEUE messages and another settting verdicts once the decision has been made.  I think this is a good pattern for NFQUEUE, we want to keep reading in pipelined fashion while we're making decisions in parallel.

With v1.9.0, this works fine (and with earlier versions too).  As of v1.11.0, this deadlocks.  

Cause appears to be #265, the "set verdict" `Send()` blocks on the reader goroutine's `Receive()` due to the change in locking.

nickgarlis:
In #265, Receive was changed from RLock to Lock to serialize concurrent Receive calls. Since Send uses RLock, it blocks for as long as Receive holds the exclusive lock. When Receive is blocking in recvmsg waiting for kernel data, Send is blocked too. This causes a deadlock in patterns like NFQUEUE, where one can goroutine read packets while another sends verdicts concurrently.

Fix by introducing a separate receiveMu that serializes concurrent Receive calls without holding mu exclusively, keeping Send and Receive concurrency compatible.

Fixes: #288

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -32,6 +32,12 @@ type Conn struct {
 	// transaction within Execute.
 	mu sync.RWMutex
 
+	// receiveMu serializes concurrent Receive and ReceiveIter calls to prevent
+	// races in multi-part message handling and the peek/allocate logic in
+	// conn_linux.go. It is separate from mu so that Send can proceed
+	// concurrently with Receive.
+	receiveMu sync.Mutex
+
 	// sock is the operating system-specific implementation of
 	// a netlink sockets connection.
 	sock Socket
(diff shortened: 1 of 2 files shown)


nickgarlis:
@florianl Do you mind having a look ?

florianl (commented):
How about a test like https://gist.github.com/florianl/13bedaf2bb26bb3c0a0ce52df9e229fb that does not depend on `time.Sleep()`?

nickgarlis:
Thanks for the thorough review @florianl! I've applied the suggested change. This pattern could be re-used for other tests too.

Issue #191, opened 2021-11-08
Prevent uint16 overflow when adding attributes via AttributeEncoder


ubiquitousbyte:
This may seem rather overanalyzed, but I was reading through the code and found a potential pitfall [in the String function of the AttributeEncoder type](https://github.com/mdlayher/netlink/blob/a81a8c23d40a0389ffcc7c9a094569ac42e6bc95/attribute.go#L634).

Isn't the following check necessary to prevent the length field from overflowing?
````go
if len(s) > math.MaxUint16-nlaHeaderLen {
    // Return error indicating an overflow
}

mdlayher:
It seems you're correct! A length check should be added to each of Bytes, Do, and String. Want to send a PR?

ubiquitousbyte:
Sure thing, mate. I'll send the PR tomorrow after work.

 I'm also seeing a small performance optimization that we can integrate without affecting code readability. 
The `AttributeEncoder` can keep track of the total (aligned) buffer length and gradually increment it whenever the caller issues an append operation (via `Bytes`, `String`, `Int8..64` etc). 

This will effectively make [this for-loop](https://github.com/mdlayher/netlink/blob/a81a8c23d40a0389ffcc7c9a094569ac42e6bc95/attribute.go#L82)
redundant, because we've already computed the length. Thus, marshalling a slice of attributes is now an `O(n)` operation, instead of `O(n^2)`. 

What do you think?

mdlayher:
Internal improvements are always welcome, as long as they don't dramatically impact the readability of the code! I'd be happy to take a look at what you have in mind.

mdlayher:
I've committed a fix for the length overflow issue, but will retitle this for the performance improvement issue.

PR #304, opened 2026-07-15
netlink: add AttributeEncoder.Len()

nickgarlis:
Add a Len() method that returns the total encoded byte size of all attributes appended so far, tracked via a push() helper.

Callers that batch attributes across multiple netlink messages can use this for pre-flight size checks. For example, nftables set elements must fit within the 16 bit limit of the nested attribute. I think this could also be useful for wgctrl-go which seems to split peers and IPs based on fixed limits [1].

push() also moves the overflow check which was previously duplicated across String, Bytes and Do. This also fixes a bug where String compared len(s) instead of len(nlenc.Bytes(s)).

Benchmarks against #191's proposal showed no significant difference, so the pre-computed length is not used when encoding. This can be revised later.

[1]: https://github.com/WireGuard/wgctrl-go/blob/a9ab2273dd1075ea74b88c76f8757f8b4003fcbf/internal/wglinux/configure_linux.go#L60

### DIFF
--- a/attribute.go
+++ b/attribute.go
@@ -478,13 +478,23 @@ type AttributeEncoder struct {
 	// If not set, the native byte order will be used.
 	ByteOrder binary.ByteOrder
 
-	attrs []Attribute
-	err   error
+	attrs  []Attribute
+	length int
+	err    error
 }
 
 // NewAttributeEncoder creates an AttributeEncoder that encodes Attributes.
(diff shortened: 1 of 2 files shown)


nickgarlis:
cc @ubiquitousbyte or @SuperQ

mdlayher (approved):

PR #310, opened 2026-07-18
netlink: extract marshalMessages for independent testing and benchmarking

nickgarlis:
SendMessages previously inlined the logic to serialize a slice of Messages into a single buffer. Extract this into a separate marshalMessages function with its own test and benchmark.

This is a first step toward investigating whether there are any benefits to optimizing the serialization path in SendMessages, as discussed in https://github.com/mdlayher/netlink/issues/254.

### DIFF
--- a/conn_linux.go
+++ b/conn_linux.go
@@ -104,18 +104,13 @@ func newConn(s *socket.Conn, config *Config) (*conn, uint32, error) {
 
 // SendMessages serializes multiple Messages and sends them to netlink.
 func (c *conn) SendMessages(messages []Message) error {
-	var buf []byte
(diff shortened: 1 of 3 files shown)


SuperQ:
The down side here is that we're going to make a copy of `messages`.

nickgarlis:
> The down side here is that we're going to make a copy of `messages`.

You mean by passing down the slice of messages into `marshalMessages` ? That should only pass the slice by reference so there is no copying unless I am missing something.

SuperQ (approved):
I don't remember if the Go compiler is smart enough to optimize away the copy. Structurally this seems fine.

Issue #278, opened 2026-04-06
Improve big-endian testing


dswarbrick:
Yet another test failure on s390x / big-endian hosts:
```
$ go test -v -count 1 -run TestNlmsgFprintf ./...
=== RUN   TestNlmsgFprintf
=== RUN   TestNlmsgFprintf/nft_add_table_ip_nat
--- FAIL: TestNlmsgFprintf (0.00s)
    --- FAIL: TestNlmsgFprintf/nft_add_table_ip_nat (0.00s)
panic: runtime error: index out of range [16] with length 16 [recovered, repanicked]

goroutine 9 [running]:
testing.tRunner.func1.2({0x3313e0, 0x4ca41d8b40c0})
        /usr/lib/go-1.26/src/testing/testing.go:1974 +0x238
testing.tRunner.func1()
        /usr/lib/go-1.26/src/testing/testing.go:1977 +0x334
panic({0x3313e0?, 0x4ca41d8b40c0?})
        /usr/lib/go-1.26/src/runtime/panic.go:860 +0x12a
github.com/mdlayher/netlink.nlmsgFprintf({0x369020, 0x4ca41d920ba0}, {{0x28, 0xa00, 0x1, 0x1, 0x4d2}, {0x4ca41d8b4090, 0x14, 0x14}}, ...)
        /home/dswarbrick/netlink/debug_linux.go:176 +0x988
github.com/mdlayher/netlink.TestNlmsgFprintf.func1(0x4ca41d9fe488)
        /home/dswarbrick/netlink/debug_linux_test.go:207 +0x7a
testing.tRunner(0x4ca41d9fe488, 0x4ca41d8ae300)
        /usr/lib/go-1.26/src/testing/testing.go:2036 +0x118
created by testing.(*T).Run in goroutine 8
        /usr/lib/go-1.26/src/testing/testing.go:2101 +0x4f8
FAIL    github.com/mdlayher/netlink     0.004s
testing: warning: no tests to run
PASS
ok      github.com/mdlayher/netlink/nlenc       0.001s [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/mdlayher/netlink/nltest      0.002s [no tests to run]
FAIL
```
Test passes as expected on amd64 / little-endian hosts.

nickgarlis:
[...]

PR #313, opened 2026-07-27
netlink: test debug output on big-endian systems

dlanov:
Part of #278.

`TestNlmsgFprintf` used captured nftables netlink messages whose
attribute headers were encoded for a little-endian system. Netlink
attribute headers use native byte order, so those fixtures produced
invalid lengths and an out-of-bounds panic when the test ran on s390x.

Build the test messages dynamically with `AttributeEncoder` instead:

- netlink attribute headers are encoded in native byte order
- nftables integer payloads are encoded in network byte order
- named `NFTA_*` and `NFT_*` constants describe the test rule
- semantic helpers construct the nested nftables expressions

This removes the runtime endian conversion and allows the existing
big-endian skip to be removed without changing production behavior.

Add s390x coverage using QEMU user-mode emulation. Rename
`TestReceiveIter` to `TestIntegrationReceiveIter` so live-kernel tests
can be excluded consistently with `-skip Integration`, while the root,
`nlenc`, and `nltest` packages run under QEMU.

Testing:

- `go test -run '^TestNlmsgFprintf$' -count=1 -v`
- `GOOS=linux GOARCH=s390x CGO_ENABLED=0 go test -exec qemu-s390x-static -tags gofuzz -count=1 -skip Integration ./nlenc ./nltest .`
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/.github/workflows/linux-test.yml
+++ b/.github/workflows/linux-test.yml
@@ -32,3 +32,33 @@ jobs:
 
       - name: Run tests
         run: go test -v -race -tags gofuzz ./...
+
+  big-endian:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version: "1.26"
+
+      - name: Check out code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install s390x emulator
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends qemu-user-static
+
+      - name: Run tests on s390x
+        env:
+          GOOS: linux
+          GOARCH: s390x
+          CGO_ENABLED: "0"
+        run: |
+          go test \
+            -exec qemu-s390x-static \
+            -tags gofuzz \
+            -count=1 \
+            -skip Integration \
+            ./nlenc ./nltest .
--- a/debug_linux_test.go
+++ b/debug_linux_test.go
@@ -4,32 +4,152 @@ package netlink
 
 import (
 	"bytes"
+	"encoding/binary"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"golang.org/x/sys/unix"
 )
 
+type debugTestAttribute func(*AttributeEncoder)
+
+func debugStringAttribute(typ uint16, value string) debugTestAttribute {
+	return func(ae *AttributeEncoder) { ae.String(typ, value) }
+}
+
+func debugUint32Attribute(typ uint16, value uint32) debugTestAttribute {
+	return func(ae *AttributeEncoder) { ae.Uint32(typ, value) }
+}
+
+func debugBytesAttribute(typ uint16, value []byte) debugTestAttribute {
+	return func(ae *AttributeEncoder) { ae.Bytes(typ, value) }
+}
+
+func debugNestedAttribute(typ uint16, attrs ...debugTestAttribute) debugTestAttribute {
+	return func(ae *AttributeEncoder) {
+		ae.Nested(typ, func(nae *AttributeEncoder) error {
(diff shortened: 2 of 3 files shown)


debug_linux_test.go
```diff
@@ -205,7 +251,9 @@ func TestNlmsgFprintf(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			w := &bytes.Buffer{}
-			nlmsgFprintf(w, tt.m, tt.colorize)
+			m := tt.m
+			m.Data = nativeEndianNftablesData(t, m.Data)
```
nickgarlis: I wonder whether it'd be easier if we changed the test array signature to something like this 
```golang
  tests := []struct {
      name     string
      m        func() Message
      colorize bool
      want     string
  }
```

and dynamically created the fixtures like so

```golang
  {
      name: "nft add table ip nat",
      m: func() Message {
          ae := netlink.NewAttributeEncoder()
          ae.String(unix.NFTA_TABLE_NAME, "nat")
          ae.Uint32(unix.NFTA_TABLE_FLAGS, 0)
          b, _ := ae.Encode()
          return Message{
              Header: Header{
                  Length:   40,
                  Type:     HeaderType(...),
                  Flags:    Request,
                  Sequence: 1,
                  PID:      1234,
              },
              Data: append(nfgenmsgHeader, b...),
          }
      },
  }
```

As long as we can trust that the attribute encoder is bug-free and well tested, we should be able to trust it to dynamically create the fixture for us. This would remove the need for the runtime conversion and would probably make failures easier to reason about.

.github/workflows/linux-test.yml
```diff
@@ -32,3 +32,42 @@ jobs:
...
+
+      - name: Check out code
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install s390x emulator
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends qemu-user-static
+
+      - name: Run architecture-independent tests on s390x
```
nickgarlis: I like the idea of qemu in user mode. It's much easier to run than the tests like this without bringing in an s390x kernel. Looks like the naming of the integration tests has not been consistent enough so this wouldn't work here

```golang
go test \
    -exec qemu-s390x-static \
    -tags gofuzz \
    -count=1 \
    -skip Integration \
    ./nlenc ./nltest .
```

nickgarlis (commented):
Hi @dlanov 

Thanks for the contribution! I've left a couple of comments on my thoughts on the PR. 

cc @aojea who implemented the mnl debugger in https://github.com/mdlayher/netlink/commit/5c0c409fd9c6abadfe140f952a3c37eb05f05108

.github/workflows/linux-test.yml
nickgarlis: If you rename `TestReceiveIter` to `TestIntegrationReceiveIter` the above command should work

debug_linux_test.go
dlanov: Thanks — updated the fixtures to construct the messages with AttributeEncoder instead of converting captured little-endian bytes at runtime. I also used named NFTA/NFT constants and semantic helpers for the nested expressions. The formatter test passes both natively and under s390x QEMU.

.github/workflows/linux-test.yml
dlanov: Thanks — renamed TestReceiveIter to TestIntegrationReceiveIter and replaced the special compile/run sequence with the suggested go test -exec qemu-s390x-static -skip Integration command for ./nlenc, ./nltest, and the root package. I verified the full command locally on s390x under QEMU.

aojea:
clearly I show this area of different architecture is not my forte :)

aojea:
ok, I see, so what I did is to put the binary format as a constant, we just need to generate the right binary format to assert on the test per platform ... I thought I screwed it more 😅 , it was just the test assertions

Thanks 

/lgtm

